PR #9, opened 2017-10-03
Updates to use React v15

gbulatov:
Updates package json to use react v.15
Fixes button text to work with react v.15


groberts314:
Going to :shipit: and then bump the package up to version `1.6.0`.

PR #14, opened 2017-11-07
Drop-down menu opens down even it does not fit the screen 

gbulatov:
#7043 Drop-down menu opens down even it does not fit the screen 

- Subscribe drop-down menu in multi-select mode to scroll events
- Hidden render of a drop-down menu to determine direction of opening the drop-down menu
- Custom styles to render drop-down menu in a correct position (re-positioning)


morethanfire:
Reviewing. Thanks!

PR #15, opened 2017-11-13
Added a prop 'allowEmpty' to be able to clear the input with min/max …

gbulatov:
Added a prop 'allowEmpty' to be able to clear the input with min/max restrictions.


src/Elements/Input.react.js
```diff
@@ -176,12 +176,14 @@ class Input extends Component {
 
     _onChange(event) {
         const { value } = this.state;
-        const { max, min, onChange } = this.props;
+        const { max, min, onChange, allowEmpty } = this.props;
         const type = this._getType();
         let newValue = event.target.value;
 
         if (type === 'number' && _.isNumber(max) || _.isNumber(min)) {
```
groberts314: These two nested if statements are pretty confusing.  Can we combine into one and simplify?

IlyaRadinsky: Fixed

gbulatov:
1. Will use required instead of allowEmpty
2. Will clean up the mess with the If statement

groberts314:
Thanks, much appreciated, sorry to be a pest about it.

morethanfire:
I pushed a fix to original codes' condition to clean it up. I am not sure what exactly the `allowEmpty` is for. What the use case?

src/Elements/Input.react.js
```diff
@@ -176,12 +176,24 @@ class Input extends Component {
...
+                }
+            } else {
+                newValue = +newValue;
+                if (_.isNumber(max)) {
+                    newValue = Math.min(max, newValue);
+                }
+                if (_.isNumber(min)) {
+                    newValue = Math.max(min, newValue);
+                }
+            }
```
groberts314: MUCH cleaner!  Thanks! :+1:   It's actually reasonable to figure out what this is doing now (for a simpleton such as myself).

@morethanfire can you take a look at latest and greatest please and make sure you're okay with it as well?

groberts314: Looks like unfortunately a conflict has to be resolved before this can be merged.  I think it can be resolved by taking this version as-is; it accomplishes the same as the fix introduced in 5b8e75b as well as introducing correct logic for the new behavior around the `required` prop.

PR #16, opened 2017-11-17
[SO] - Update max age delay

gbulatov:
BI 7179 [SO] - Update max age delay

* adds 500ms delay before validation
* updates documentation

### DIFF
--- a/src/Elements/Input.react.js
+++ b/src/Elements/Input.react.js
@@ -214,7 +214,7 @@ class Input extends Component {
             } else {
                 this.setState({ value: newValue });
             }
-        }, 300);
+        }, 500);
         this.setState({ value: newValue });
     }
 
--- a/src/Elements/Input.react.js
+++ b/src/Elements/Input.react.js
@@ -188,12 +188,12 @@ class Input extends Component {
         const type = this._getType();
         let newValue = event.target.value;
 
-        if (this.inputTimer) {
-            clearTimeout(this.inputTimer);
-        }
+        if (type === 'number') {
+            if (this.inputTimer) {
+                clearTimeout(this.inputTimer);
+            }
 
-        this.inputTimer = setTimeout(() => {
-            if (type === 'number') {
+            this.inputTimer = setTimeout(() => {
                 if (_.isEmpty(newValue)) {
                     if (required) {
                         newValue = _.isNumber(min) ? min : (_.isNumber(max)? max : 0);
@@ -207,15 +207,21 @@ class Input extends Component {
                         newValue = Math.max(min, newValue);
                     }
                 }
-            }
+
+                this.setState({ value: newValue });
+
+                if (!_.isUndefined(onChange)) {
+                    onChange(newValue);
+                }
+            }, 500);
+            this.setState({ value: newValue });
+        } else {
+            this.setState({ value: newValue });
 
             if (!_.isUndefined(onChange)) {
                 onChange(newValue);
-            } else {
-                this.setState({ value: newValue });
             }
-        }, 500);
-        this.setState({ value: newValue });
+        }
     }
 
     _onClick(event) {
--- a/src/Elements/Input.react.js
+++ b/src/Elements/Input.react.js
@@ -22,11 +22,19 @@ class Input extends Component {
         this._onClick = this._onClick.bind(this);
         this._onFocus = this._onFocus.bind(this);
         this._onKeyDown = this._onKeyDown.bind(this);
+
+        this.inputTimer = null;
     }
 
     componentWillReceiveProps(nextProps) {
         if (this.props.value !== nextProps.value) {
             this.setState({ value: nextProps.value });
+        }
+    }
+
+    componentWillUnmount() {
+        if (this.inputTimer) {
+            clearTimeout(this.inputTimer);
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Elements/Input.react.js
```diff
@@ -180,27 +188,34 @@ class Input extends Component {
...
-        } else {
-            this.setState({ value: newValue });
-        }
+            if (!_.isUndefined(onChange)) {
+                onChange(newValue);
+            } else {
+                this.setState({ value: newValue });
+            }
+        }, 500);
+        this.setState({ value: newValue });
```
groberts314: Just asking a clarifying question to ensure I understand: If an initially invalid value is typed, it looks like the immediate `setState()` will allow that value to be rendered, and then the logic in the timeout will kick in and possibly do another `setState()` with the "corrected" value (e.g. if the value was outside of the min...max range)?  That seems like correct behavior.

gbulatov: Yes, exactly. So if we have a valid range of 25 to 90, and trying to type 40, the first digit 4 is less than 25, and before it was immediately checked against Min value of 25 and corrected it. Similar thing for Max value. Now we just allow user to type all digits before validation.

groberts314: Yup, makes total sense.  Thanks for supplying a concrete example there. :+1: 

I am in agreement with @morethanfire though ... we probably want to ensure that the new behavior we're introducing with the timeout/delay to only applies when `type === 'number'` ... right?  No need to make `<Input>` elements of other types delay firing `onChange`, right?

Maybe something like this...?

```JavaScript
if (type === 'number') {
    // For numeric inputs, setup a timeout delay to apply validation and possible adjustment to the new value
    // followed by `onChange` if it's defined
    this.inputTimer = setTimeout(() => {
        if (_.isEmpty(newValue)) {
            if (required) {
                // If a value is required but was not supplied then default to either the configured minimum value or 0
                newValue = _.isNumber(min) ? min : 0;
            } /* else { newValue = null; } // should we also do this? To ensure it's `null` and not, e.g. an empty string? */
        } else {
            newValue = +newValue;

            if (_.isNumber(max)) {
                newValue = Math.min(max, newValue);
            }
            
            if (_.isNumber(min)) {
                newValue = Math.max(min, newValue);
            }
        }

        this.setState({ value: newValue });
        
        if (!_.isUndefined(onChange)) {
            onChange(newValue);
        }
    }, 500);

    // Then set state with raw (un-validated) new value immediately, so that we re-render to echo the supplied value
    this.setState({ value: newValue });
} else {
    // For all other types besides number, follow the normal pattern:
    // set state and then fire `onChange` immediately (no delay)
    this.setState({ value: newValue });
    
    if (!_.isUndefined(onChange)) {
        onChange(newValue);
    }
}
```

Possibly it's not so big a deal with a delay of only 500 ms, but Cameron's thought was to try and avoid changing the current behavior of other input types of we don't have to.  Let us know if you and Ilya concur that this makes sense.

We don't necessarily have to rush this, but since it is a bit broken in the release branch and on CM-Test at the moment, it'd be nice to fix Monday in the release branch.  Should be do-able since Monday in Russia happens before Monday in USA, right? ;-)  If it doesn't happen Monday; it's okay, not a big deal ... we'll fix it in dev and it'll get released in due time.

Thanks gents.

gbulatov:
Updated version according to the comments. Maybe we need to add `default` in the future

groberts314:
Changes look okay to me.  @morethanfire ?

src/Elements/Input.react.js
```diff
@@ -207,15 +207,21 @@ class Input extends Component {
                         newValue = Math.max(min, newValue);
                     }
                 }
-            }
+
+                this.setState({ value: newValue });
```
morethanfire: `this.setState({ value: newValue }); needs to be in an else if `onChange` is undefined.

src/Elements/Input.react.js
```diff
@@ -207,15 +207,21 @@ class Input extends Component {
...
+
+                if (!_.isUndefined(onChange)) {
+                    onChange(newValue);
+                }
+            }, 500);
+            this.setState({ value: newValue });
+        } else {
+            this.setState({ value: newValue });
```
morethanfire: this.setState({ value: newValue }); needs to be in an else if onChange` is undefined.

src/Elements/Input.react.js
groberts314: Just to confirm ... it _doesn't_ need to set its own internal state if the `onChange` prop is specified?  I guess we're relying on the consuming component to pass the new value in via props (i.e. the consuming component sets some of its own state or whatever is bound to the `value` prop of the `<Input>` in its function that handles `onChange`)?

morethanfire: Yes, exactly.

IlyaRadinsky: Fixed

src/Elements/Input.react.js
IlyaRadinsky: Fixed

src/Elements/Input.react.js
```diff
@@ -181,25 +189,41 @@ class Input extends Component {
...
+                } else {
+                    onChange(newValue);
                 }
-            }
-        }
+            }, 500);
 
-        if (!_.isUndefined(onChange)) {
-            onChange(newValue);
+            if (_.isUndefined(onChange)) {
```
morethanfire: Do we need lines 218-220 if we have 211-215?

IlyaRadinsky: Yes, we do. Lines 218-220 immediately change the state, but lines 211-215 change the state by timer after the validation. These states can be different, and this is the main reason for using the timer - allows the user to quickly enter text (maybe not valid at the moment), but perform the validation after a pause.

morethanfire: Sounds good and looks good. Thanks!

groberts314: I concur ... that was the _whole reason_ for this PR. :-)  It all has to do with the Min and Max Age inputs when setting Ministry or Serving Opportunity eligibility requirements.

PR #21, opened 2018-03-20
BUG #7895 [Every] Spin control: spin buttons changes values, but doesn’t save them on Save button

gbulatov:
Invoke callback on press spinner buttons


morethanfire:
👍

PR #24, opened 2018-06-14
Adding AutoFocus for Textarea and Input.

mdjacobs:
Adding props for AutoFocus

### DIFF
--- a/docs/src/js/components/ElementsTextArea.react.js
+++ b/docs/src/js/components/ElementsTextArea.react.js
@@ -38,6 +38,20 @@ export default class AutoHeightSample extends React.Component {
 
 }`;
 
+const autoFocusSample = `import React from 'react';
+
+import TextArea from 'components/UI/Elements/TextArea.react';
+
+export default class AutoFocusSample extends React.Component {
+
+    render() {
+        return (
+            <TextArea autoFocus={true} />
+        );
+    }
+
+}`;
+
 const disabledSample = `import React from 'react';
 
 import TextArea from 'components/UI/Elements/TextArea.react';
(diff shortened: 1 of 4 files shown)


src/Elements/Input.react.js
```diff
@@ -182,6 +183,15 @@ class Input extends Component {
                 this.setState({ inputActionsTopPosition: inputTop });
             }
         }
+
+        if(autoFocus) {
+            let len = this.state.value.length;
+            this.state.value[len-1].focus();
```
groberts314: What if length is 0?

docs/src/js/components/ElementsTextArea.react.js
```diff
@@ -492,6 +506,20 @@ export default class ElementsTextArea extends React.Component {
                     {autoHeightSample}
                 </Highlighter>
 
+                {/* Auto Focus */}
+                <Header size="large" style={{ marginTop: '55px' }} sub={true}>
```
morethanfire: Let's remove `={true}` from the `sub` prop.

docs/src/js/components/ElementsTextArea.react.js
```diff
@@ -492,6 +506,20 @@ export default class ElementsTextArea extends React.Component {
...
+                </Header>
+
+                <TextArea autoFocus={true} />
+
+                <Highlighter customStyle={{ marginBottom: '44px', marginTop: '44px' }}>
+                    {autoFocusSample}
+                </Highlighter>
+
                 {/* Disabled */}
                 <Header size="large" style={{ marginTop: '55px' }} sub={true}>
```
morethanfire: Same thing here: remove `={true}` from the `sub` prop.

src/Elements/Input.react.js
```diff
@@ -182,6 +183,15 @@ class Input extends Component {
                 this.setState({ inputActionsTopPosition: inputTop });
             }
         }
+
+        if(autoFocus) {
```
morethanfire: Need a space between `if` and `(`.

src/Elements/Input.react.js
```diff
@@ -182,6 +183,15 @@ class Input extends Component {
...
+            let len = this.state.value.length;
+            this.state.value[len-1].focus();
+
+            this.setState({
+                isFocused: true
+            })
+        }
```
morethanfire: I'm wondering if this should just be set in the `constructor()` instead of having the component render again?

mdjacobs: It does help with updating the classes that are on the components.

docs/src/js/components/ElementsTextArea.react.js
mdjacobs: So just pass `sub`?

PR #26, opened 2018-07-24
Task [10087] Immersive nav (Tier 3) -- part of PBI [8769] Serving Opportunities 1.1

gbulatov:
* Adds an option for a drawer to open from a left side (also support second wing and other drawer features)
* Updates documentation

### DIFF
--- a/docs/src/js/components/ModulesDrawer.react.js
+++ b/docs/src/js/components/ModulesDrawer.react.js
@@ -48,6 +48,45 @@ export default class DrawerSample extends React.Component {
 
 }`;
 
+const leftDrawerSample = `import React from 'react';
+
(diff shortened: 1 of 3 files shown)


gbulatov:
@morethanfire @groberts314 Please have a look, we need this for tier 3 navigation in SO 1.1

morethanfire:
I am struggling with the the term `side` vs `position`. For example in the **Rail** component we use position. I'd like to be consistent with the term we use. Are you opposed to updating the update to use `position`?

IlyaRadinsky:
OK, will do tomorrow

morethanfire:
Thanks!

morethanfire:
👍 
Love it! I will merge this and kick out a release today.

PR #29, opened 2018-09-12
Added color to info bar

bittu1028:


morethanfire:
👍 I totally didn't know what to call these background gradient colors, so numerical characters they got. Saved me so much time not having to be clever.

PR #31, opened 2018-10-24
Tweak `Image` component and its documentation

groberts314:
* Fix it so that if `name` and `src` props are both specified (in `avatar` mode), it only renders the image and not the initials.
* Fix up the documentation page so that (hopefully :crossed_fingers: ) it shows the Marty McFly image correctly when deployed (where the path is different and a bit cleaner).
* Also add a sample to the documentation page to test `name` + `src` use case.

### DIFF
--- a/docs/src/js/components/ElementsImage.react.js
+++ b/docs/src/js/components/ElementsImage.react.js
@@ -21,7 +21,7 @@ export default class ImageSample extends React.Component {
 
     render() {
         return (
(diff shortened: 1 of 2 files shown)


groberts314:
@morethanfire if this meets with your approval I'd like to merge it and release it as `1.16.2`.  Thank you sir!

morethanfire:
👍

groberts314:
Cool deal ... I'll merge this sucka and do the release management for 1.16.2.  Thanks for the review & thumbs up.

PR #33, opened 2018-10-27
Make Comment component "editable"

groberts314:
Add ability to edit and delete comments using the Comment component
Being done in service of Workflows 1.0 User Task Dashboard
[PBI 11063](https://saddlebackchurch.visualstudio.com/DefaultCollection/Church%20Management/_workitems/edit/11063): Ability to leave a comment on a task
[Task 12380](https://saddlebackchurch.visualstudio.com/DefaultCollection/Church%20Management/_workitems/edit/12380): BUG: My Dashboard > Task Details Drawer: Cannot edit/delete comments
Sketch: https://sketch.cloud/s/ORzY2/Pr3OEv
Requirements: https://saddleback.egnyte.com/dl/YoLk2mpHTO

Let's please use this PR as a discussion forum to dial the basic UI/UX of the Comment component.  Then we'll work on the specific placement of the component in the My Dashboard Task Drawer.

* Comment author can edit and/or delete own comments
* Task supervisor can delete anyone's comments

The React CM UI library component will not attempt to make any assumptions about these kinds of "permissions" -- instead the consumer sets props indicating what actions are and are not allowed.

image

image

image

(Banner is not baked in -- it is just for illustrative purposes on the documentation page.  Component simply calls an `onSaveEdit()` prop function provided by the consumer, allowing the consumer to issue API calls, show banners, whatever to save the comment and inform the user it's done.)

### DIFF
--- a/docs/src/js/components/ElementsComment.react.js
+++ b/docs/src/js/components/ElementsComment.react.js
@@ -31,8 +31,8 @@ export default class CommentSample extends React.Component {
         super(props);
 
         this.state = {
-            editableCommentText: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut pretium pretium tempor.',
-            editableComment2Text: 'Pellentesque sagittis quam enim, a euismod nisl tristique ac. Duis rutrum accumsan nisl sit amet congue. Vivamus ut convallis velit, nec adipiscing mi. Sed semper dui ut velit eleifend tincidunt.',
+            editableCommentText: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+            editableComment2Text: 'Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.',
             isRemoveBannerOpen: false,
             isSaveBannerOpen: false
         };
@@ -58,14 +58,25 @@ export default class CommentSample extends React.Component {
 
                 <Comment
                     canDelete
-                    canEdit
+                    canEdit={false}
                     detailsPosition="right"
                     isEditable
                     name="Jessica Jones"
                     onDelete={this._onRemoveComment}
+                    time={1536941520}
+                >
+                    {'Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.'}
+                </Comment>
+
+                <Comment
+                    canDelete
+                    canEdit
+                    isEditable
+                    name="Joe Smith"
+                    onDelete={this._onRemoveComment}
                     onSaveEdit={this._onSaveComment2}
                     text={editableComment2Text}
-                    time={1536941520}
+                    time={1536941640}
                 >
                     {editableComment2Text}
                 </Comment>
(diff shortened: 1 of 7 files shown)


src/Elements/Comment.react.js
```diff
@@ -38,16 +57,111 @@ class Comment extends Component {
...
         );
     }
 
+    _onCancelEditClick() {
+        this.setState({ isEditMode: false });
+    }
+
+    _onCommentChange(value) {
+        this.setState({ updatedCommentText: value });
+    }
```
groberts314: @morethanfire ... I wasn't able to get the Text Area to work here as an "uncontrolled" component.  I needed to initialize the Text Area's value and then do the standard `onChange` deal.  Note that I'm holding the "dirty" unsaved but updated comment text in state in here.  Maybe you and I can play with it Monday and make it better.

src/Elements/Comment.react.js
```diff
@@ -64,11 +178,24 @@ class Comment extends Component {
...
+    canEdit: PropTypes.bool,
+    canRemove: PropTypes.bool,
     className: PropTypes.string,
     detailsPosition: PropTypes.oneOf([ 'left', 'right' ]),
+    isEditable: PropTypes.bool,
     name: PropTypes.string,
+    onRemove: PropTypes.func,
+    onSaveEdit: PropTypes.func,
     style: PropTypes.object,
+    text: PropTypes.string,
```
groberts314: @morethanfire For edit mode, I had to expose a prop where the contents on the comment come in as a string (whereas the previous read-only version used `children` prop to accept a JSX node, which is better and more flexible.  Is there a way to render JSX to a string and avoid having to do this?  Any thoughts?

src/Elements/Comment.react.js
```diff
@@ -38,16 +57,111 @@ class Comment extends Component {
...
+                                    <a
+                                        className="save-link"
+                                        disabled={!canEdit}
+                                        onClick={this._onSaveClick}
+                                        style={{ marginRight: '11px' }}
+                                    >
+                                        {'Save'}
+                                    </a>
+                                    <a
+                                        className="remove-link"
```
groberts314: I added classes like these so consumers can apply styles.  For instance, in the comps (https://sketch.cloud/s/ORzY2/Pr3OEv) the "Remove" CTA is black versus "Edit" and "Save" being our standard link color.  I didn't want to bake that in ... I was thinking rather that consumer can apply CSS to do these kinds of customizations.  Is this a good approach?  Anything else in here need CSS classes?  Do you think we need props anywhere to be able to apply additional classes (like we do elsewhere in the React CM UI library)?

docs/src/js/components/ElementsComment.react.js
```diff
@@ -96,7 +195,72 @@ export default class ElmentsComment extends React.Component {
...
+                    id="remove-success"
+                    level="success"
+                    isOpen={isRemoveBannerOpen}
+                    message="Comment Removed!"
+                    onClose={this._onCloseRemoveBanner}
+                    title="Comment Removed"
+                    type="notification"
+                />
+
+                <Banner
```
morethanfire: How are the Banners going to be rendered, location-wise, in the UI?

groberts314: On the UI docs page?  Should be up near the top of the view port as normal.  I can demo this live if you'd like.  Just to reiterate, I only wired up banners for demo purposes on the documentation page.  That is 100% external to the component itself ... it's 100% up to the consumer (as it should be in my opinion).

groberts314:
I'm going to go ahead and :shipit: 
This is going to go into a `1.17.0` release for React CM UI.
If there are any issues, I'll get them resolved and follow up with one or more `1.17.x` patch releases as needed.

groberts314:
Merging this for reals! :-P

PR #36, opened 2019-01-18
[dev] [part of] PBI [14321] Add DurationPicker component (will be used for Follow Up (Task) Due Date policy)

groberts314:
**Overview**
Adds a **`DurationPicker`** component to the React CM UI library.  This component allows you to specify a length of time by various units (days, hours, minutes, etc.).  As an example, this component will be used to say that Follow Up Tasks (for the "Workflows" feature) are due after X number of days or Y number of hours.

**Background**
As a part of cleaning up some technical debt in the Workflows area (mostly in API/back-end but also some in front-end) I'm going to try and decompose the "Workflow Rule Editor" component into some smaller sub-components (the component in question currently lives alongside other components for the editing of Connection Card [Response Card] Templates).  Currently, something very much like the `DurationPicker` component here exists inside a much larger and complicated component; it seemed like a good first candidate to extract out.

image

### DIFF
--- a/docs/src/js/components/CoreApp.react.js
+++ b/docs/src/js/components/CoreApp.react.js
@@ -335,6 +335,14 @@ let routes = (
                 }}
             />
             <Route
+                path="duration-picker"
+                getComponent={(location, callback) => {
+                    require.ensure([], require => {
+                        callback(null, require('components/ModulesDurationPicker.react'))
+                    }, 'ModulesDurationPicker')
+                }}
+            />
+            <Route
                 path="modal"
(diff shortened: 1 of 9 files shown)


groberts314:
Thanks @IlyaRadinsky !  @morethanfire ... would love some of your review if you have some time, especially on any markup and styling aspects here.

morethanfire:
👍 Very cool!
@groberts314, do you want to merge and cut a release?

groberts314:
Yessir I can take care of that.  Thanks for the review.

PR #38, opened 2019-02-25
Added light font for source sans sarif

bittu1028:


groberts314:
Please forgive my ignorance ... but what we're doing here is making a couple of additional font weight variations available?  Seems okay to me.  Seems additive, which is generally safe.  @morethanfire ... any objections?

groberts314:
@morethanfire Ping! :-)
@bittu1028 Can you advise on what feature (Backlog Item, Bug etc.) this affects, and the urgency?

bittu1028:
@groberts314 I need the light font in the title of quick notification drawer which was posted as a bug by James in sketch cloud.

groberts314:
@bittu1028 Let's linger on this during the 10:00 AM scrum with @morethanfire and James and such to see if this many font weight variations are really needed.

groberts314:
Per the linger ... drumroll please ... :shipit: 🎉 
(Was hoping that if we _did_ end up shipping this it rolled with `1.18.6` but this will have to go into `1.18.7` I suppose ... sigh)

PR #41, opened 2019-04-09
Upgraded React-Portal and React-Dnd

morethanfire:


docs/src/js/components/ModulesDrawer.react.js
```diff
@@ -975,6 +974,7 @@ export default class ModulesDrawer extends React.Component {
     }
 
     _onPathDrawerClick(path) {
+        console.log('path', path);
```
groberts314: If this is a useful `console.log()` message on the documentation page for demonstrating functionality, should we put the the ESLINT suppression comment?  If this was for temporary debugging only, let's please remove it altogether.

src/Modules/Drawer.react.js
```diff
@@ -129,6 +131,23 @@ class DrawerHeader extends Component {
...
+    }
+
+    _onHeaderUpdate(node) {
+        const headerEl = ReactDOM.findDOMNode(this);
+
+        if (headerEl) {
+            const paddingTop = headerEl.offsetHeight;
+
+            if (paddingTop !== this._drawerContainerInnerPaddingTop) {
+                this._drawerContainerInnerPaddingTop = `${paddingTop + 27}px`;
```
groberts314: Is it maybe worth defining a constant for 27 as the default padding?  In case next week design team takes a fit and wants to update it to 33 or some other value; that way you only have to update it in one place not two places (here and up on line 109).  Just a thought.

src/Modules/Drawer.react.js
```diff
@@ -223,11 +255,11 @@ class Drawer extends Component {
...
-                    closeButton: closeButton,
-                    inverse: inverse,
-                    onClose: onClose,
-                    title: title,
-                    titleTruncate: titleTruncate
+                    closeButton: this.props.closeButton,
+                    inverse: this.props.inverse,
+                    onClose: this._onClose,
+                    title: this.props.title,
+                    titleTruncate: this.props.titleTruncate,
```
groberts314: We cannot use the props value we destructured above?

src/Modules/Modal.react.js
```diff
@@ -144,17 +183,17 @@ class Modal extends Component {
...
-                                    closeButton: closeButton,
-                                    inverse: inverse,
-                                    onClose: onClose,
-                                    title: title,
-                                    titleTruncate: titleTruncate
+                                    closeButton: this.props.closeButton,
+                                    inverse: this.props.inverse,
+                                    onClose: this._onClose,
+                                    title: this.props.title,
+                                    titleTruncate: this.props.titleTruncate,
```
groberts314: Same ... is there a reason we cannot use the destructured ones?

docs/src/js/components/ModulesDrawer.react.js
morethanfire: It was/is to show something happened to the dev.

groberts314: Yep, so that is a useful log message, on a documentation page.  So let's keep it and exempt it from ESLINT, right?

morethanfire: Nope, I was wrong, totally not needed.

src/Modules/Drawer.react.js
morethanfire: Good idea. Might as well implement it.

src/Modules/Drawer.react.js
morethanfire: done.

src/Modules/Modal.react.js
morethanfire: Done.

src/Modules/Drawer.react.js
```diff
@@ -257,23 +257,23 @@ class Drawer extends Component {
...
                 return React.cloneElement(child, {
-                    closeButton: this.props.closeButton,
-                    inverse: this.props.inverse,
+                    closeButton: closeButton,
+                    inverse: inverse,
                     onClose: this._onClose,
-                    title: this.props.title,
-                    titleTruncate: this.props.titleTruncate,
+                    title: title,
+                    titleTruncate: titleTruncate,
```
groberts314: If I were getting _super_ nit-picky, I'd suggest we can use the short-hand whenever assigning a value to a field of the same name.  e.g. This can be written as:
```JavaScript
return React.cloneElement(child, {
    closeButton,
    inverse,
    title,
    titleTruncate,
    onClose: this._onClose,
});
```

But, meh, we don't have to go that far. :-)

src/Modules/Modal.react.js
```diff
@@ -183,11 +185,11 @@ class Modal extends Component {
...
                                 {header ? React.Children.map(this.props.children, c => React.cloneElement(c, {
-                                    closeButton: this.props.closeButton,
-                                    inverse: this.props.inverse,
+                                    closeButton: closeButton,
+                                    inverse: inverse,
                                     onClose: this._onClose,
-                                    title: this.props.title,
-                                    titleTruncate: this.props.titleTruncate,
+                                    title: title,
+                                    titleTruncate: titleTruncate,
```
groberts314: Similar here ... we can short-hand this ever further if we really want, but that's definitely in super nit-pick land.

PR #47, opened 2019-05-20
[dev] [NoPBI] TO BE DISCUSSED: Migration to gulp4

gbulatov:
We use Node 12 internally, so we were forced to use Gulp4 (gulp3 does not support node12 correctly)
requires `npm install`


morethanfire:
I am still on node v.10.8.0. Do we know that the upgrade of gulp will work with an older version of node?

IlyaRadinsky:
@morethanfire  Should work, but not sure. Could you test it locally, please?

morethanfire:
When I get some down time I will give it a test. Until then, I am keeping this PR open.

IlyaRadinsky:
👍 
No rush to merge, thanks.

groberts314:
I'm wondering if this effort can be part of a broader UI technical debt item in Sprint 45.
* Get both `react-cm-ui` and Healthy Church on the "latest and greatest" version of Webpack and Gulp (or at least some reasonably up-to-date, "stable" version that we determine, hopefully the _same_ version for both repos/projects)
* See if we can address at least a subset of other package deprecation and/or security vulnerability warnings we see during `npm install`, especially on the build server.

@JLeeC @morethanfire @gbulatov @IlyaRadinsky @Transmuter ... thoughts?

morethanfire:
I am all for this suggestion. I believe getting off Webpack v3 and upgrading to v4 would be a good decision. Gulp isn't as critical/important - maybe I am not aware of something? - but it doesn't hurt to update at the same time.

PR #52, opened 2019-08-23
[release/5.1.1] AB#19803 Person Record 2.0 UI framework & structure

fantaJinMode:
Implementation of `Drawer.Details` same as `Page.Details`

NOTE: 
Also consists fixes of typo, Case Sensitive folder name shows problem on Linux

### DIFF
--- a/src/elements/pageFiltersDrawer.scss
+++ b/src/elements/pageFiltersDrawer.scss
@@ -1,0 +1,33 @@
+@import '../shared/styles/colors';
+
+.ui.page--filters_drawer {
+    &.ui.drawer .drawer-container-inner {
+        .drawer-close-button-container {
+            top: 16px;
+        }
+        .drawer-header {
+            background-color: $bkgd-light;
+            padding-top: 49px;
+            .title {
+                top: 16px;
+            }
+        }
+    }
+}
+
+@media only screen and (max-width: 767px) {
+    .ui.page--filters_drawer {
+        .drawer-container-inner {
+            padding-left: 11px !important; // sass-lint:disable-line no-important
+            padding-right: 11px !important; // sass-lint:disable-line no-important
+        }
+        .drawer-header .title {
+            padding-left: 11px !important; // sass-lint:disable-line no-important
+            padding-right: 11px !important; // sass-lint:disable-line no-important
+        }
+        .drawer-close-button-container {
+            right: 11px !important; // sass-lint:disable-line no-important
+            width: 16px !important; // sass-lint:disable-line no-important
+        }
+    }
+}
--- a/src/modules/drawerDetails.js
+++ b/src/modules/drawerDetails.js
@@ -1,9 +1,8 @@
-import _ from 'lodash';
 import Details from './details';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-const PageDetails = (props) => {
+const DrawerDetails = (props) => {
     const {
         bleed,
         className,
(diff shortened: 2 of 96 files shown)


src/elements/pageFiltersDrawer.scss
```diff
@@ -0,0 +1,33 @@
...
+            padding-left: 11px !important; // sass-lint:disable-line no-important
+            padding-right: 11px !important; // sass-lint:disable-line no-important
+        }
+        .drawer-header .title {
+            padding-left: 11px !important; // sass-lint:disable-line no-important
+            padding-right: 11px !important; // sass-lint:disable-line no-important
+        }
+        .drawer-close-button-container {
+            right: 11px !important; // sass-lint:disable-line no-important
+            width: 16px !important; // sass-lint:disable-line no-important
```
groberts314: Lots of `!important`s here.  As I understand it, that's a good thing to flag as a potential code smell, and the mechanism of last resort.  So, I'm dutifully calling it out.  If there is a good case that `!important` really is required here, so be it.  Just want to ensure that everyone (mainly @morethanfire ) is aware of the reason why and is on board with the usage.

fantaJinMode: @groberts314 I haven't worked on this file. Might be automatic es-lint fixing while saving it. @morethanfire might have idea about it.

groberts314: Interesting.  It does appear this file is already like this in `dev` branch.
I think this PR would be introducing this file to `release/5.1.1` branch where it doesn't exist yet.  That's why it shows up as diffed.  My bad.  Then more than likely Señor Brewer is already well aware of the `!important`s. 😁

src/modules/drawerDetails.js
```diff
@@ -0,0 +1,42 @@
...
+                bleed={bleed}
+                className={className}
+                color={color}
+                columnProps={columnProps}
+                columns={columns}
+                data={data}
+                style={style}
+                moduleType="drawer"
+            />
+        </React.Fragment>
```
morethanfire: No need for `<React.Fragment />`, let's remove. Also, let's spread the passed in props instead of listing them this way. 

e.g. 
```
<Details
    {...props}
    moduleType="drawer"
/>
```

src/elements/pageFiltersDrawer.scss
morethanfire: These `!important`s are good.

src/modules/pageDetails.js
```diff
@@ -1,156 +1,33 @@
...
+        className,
+        color,
+        columnProps,
+        columns,
+        data,
+        style,
+    } = props;
+
+    return (
+        <React.Fragment>
```
morethanfire: Same comment in drawerDetails file for this file.

src/modules/pageDetails.scss
```diff
@@ -36,7 +36,7 @@ $divider-color: rgba(255, 255, 255, .25);
 
 @media only screen and (min-width: 768px) {
     .ui.page--details {
-        &.page--details-bleed {
+        &.details-bleed {
             margin: 0 -22px;
         }
         .ui.info-bar {
```
morethanfire: Let's strip some of the common css out of this file and make a `details.scss` file so that it only applies to the two different components.

src/elements/pageFiltersDrawer.scss
groberts314: I guess they're, err, _important_ eh? 🤣

src/modules/drawerDetails.js
fantaJinMode: Fixed

src/modules/pageDetails.js
fantaJinMode: Fixed

src/modules/pageDetails.scss
fantaJinMode: For this, Just need a little more clarification. Striping out  common css from `pageDetails.scss` & `drawerDetails.scss` and placing it into `details.scss`. Here's what I am thinking, need to add new common class which would work for both page and drawer component like below
`className={`${moduleType}--details-columns-container details-columns-container`}`
Here, `details-columns-container` is new common class. I hope you are getting my idea. Am I thinking correctly please suggest?

fantaJinMode: Fixed

fantaJinMode:
Thanks, Just missed it, working on those conflicts.

PR #56, opened 2019-09-11
Fixes generating the key for the Page.Table header cell

IlyaRadinsky:
The `column.header` could be not only text (e.g. it maybe the `Checkbox` or `div`), so, the header cell key should be generated by index.

### DIFF
--- a/src/modules/pageTable.js
+++ b/src/modules/pageTable.js
@@ -93,7 +93,7 @@ class PageTable extends React.PureComponent {
                                 return (
                                     <Table.HeaderCell
                                         className="page--table_header_cell"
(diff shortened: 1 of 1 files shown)


IlyaRadinsky:
@morethanfire please review.

groberts314:
Looks super simple and straight-forward to me.  :shipit:

groberts314:
@IlyaRadinsky maybe you and @morethanfire (and myself if you gents need or want my participation as well) can pow-wow regarding "release planning" for React CM UI.  Obviously sometimes there will be unplanned bug fixes, but if there is going to be a substantial amount of planned work (say for My Ministry Roster Mobile or Person Record 2.0) then let's try and plan it a bit.

PR #58, opened 2019-09-17
[dev] PBI AB#20206 [My Ministry > Members] [mobile] - Convert table to new mobile table

IlyaRadinsky:
sticky-columns

The main idea: when we need to have the sticked columns, we render two tables - the first with horizontally scrollable table, the second - with sticked columns. This trick allows us to scroll the table as by vertically, as by horizontally. The column sizes synchronized on Resize event.

New props:
    *minWidth*: min scrollable table width
    *splitter*: optianal Splitter icon and resizing logic
    *stickyColumnWidth*: the min width of sticked column
    *stickyColumns*: the number of sticked columns (can be more then 1)

@morethanfire have a look, please.


src/modules/pageTable.js
```diff
@@ -12,12 +14,17 @@ class PageTableRow extends React.PureComponent {
...
+            rowIndex,
+            splitter,
+        } = this.props;
+        const sizes = this.props.sizes || [];
 
         return (
-            <Table.Row
-                onClick={isClickable ? this._onClick : null}
-            >
+            <Table.Row onClick={this._onClick}>
```
groberts314: Did we lose `isClickable` check here deliberately?  I guess you're thinking that it is handled in `_onClick()` so no need to conditionally bind a function or not here?

src/modules/pageTable.js
```diff
@@ -27,10 +34,26 @@ class PageTableRow extends React.PureComponent {
                         accessor = column.accessor(row);
                     }
 
+                    const style = {};
+                    const size = (sizes[rowIndex] || [])[index];
```
groberts314: The ` || []` fallback and then indexing it seems a bit odd to me as clearly that would result in `undefined`.  That's probably perfectly OK ... one of those instances where JavaScript allows you to do things that would blow up loudly in C# so it messes with my head a bit. =)  Probably there nothing to do here, but I'm just flagging it for discussion, in case there is a "nicer" way to default this if `sizes` 2-d array is undefined or empty or what-not.

src/modules/pageTable.js
```diff
@@ -27,10 +34,26 @@ class PageTableRow extends React.PureComponent {
...
+
+                    if (idPrefix === 'column') {
+                        if (splitter && _.last(columns) === column) {
+                            style.borderRight = '1px solid #edf1f5';
+                        }
+                    }
+
                     return (
                         <Table.Cell
+                            id={`tableCell-${idPrefix}-${rowIndex}-${index}`}
```
groberts314: ID here should probably be `table-cell...` rather than `tableCell` ... we seem to favor a combination of "snake case" and "kebab case" for our IDs and classes rather than "camel case".  But I'll leave the final decision on that to yourself and Mssr. @morethanfire .

src/modules/pageTable.js
```diff
@@ -41,36 +64,53 @@ class PageTableRow extends React.PureComponent {
     }
 
     _onClick() {
-        const { isClickable, row, rowProps } = this.props;
+        const { isClickable, onClick, row, rowProps } = this.props;
         const isTextSelect = window.getSelection().toString();
 
-        if (isClickable && !isTextSelect) {
+        if (_.isFunction(onClick)) {
```
groberts314: So no `isClickable` check for this case?  Basically, if the consumer says `isClickable={false}` _and_ binds an `onClick` handler function, you're assuming that the binding of the function indicates intention more than `isClickable={false}`? =)

groberts314:
I'll also do another shout-out here to inclusion of the animated GIF.  Very helpful and it looks awesome; I'm looking forward to seeing this baby in action in the HC Back-Office application very soon. 👍

morethanfire:
Hey, sorry @IlyaRadinsky I was unable to get to this PR yesterday. I will be getting to it shortly here this morning.

src/modules/pageTable.js
morethanfire: Let's do `table--table_cell-${idPrefix}_${rowIndex}_${index}`.

src/modules/pageTable.js
```diff
@@ -41,36 +64,53 @@ class PageTableRow extends React.PureComponent {
...
 
 PageTableRow.propTypes = {
     columns: PropTypes.array.isRequired,
+    idPrefix: PropTypes.string.isRequired,
     isClickable: PropTypes.bool,
+    onClick: PropTypes.func,
     row: PropTypes.object.isRequired,
+    rowIndex: PropTypes.number.isRequired,
     rowProps: PropTypes.func,
+    sizes: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)),
```
morethanfire: Nice!

src/modules/pageTable.js
```diff
@@ -119,23 +175,201 @@ class PageTable extends React.PureComponent {
...
+                            columns={_.slice(columns, 0, stickyColumns)}
+                            idPrefix="column"
+                            sizes={sizes}
+                            onSplitter={this._onSplitterClick}
+                        />
+                    </div>
+                </div>
+            );
+        } else {
+            return <PageTable {...this.props} />;
```
morethanfire: Let's bring the `<div className="page--table" />` down to this parent component and wrap this `<PageTable />` child component and the above stickyColumns jsx.

`class UITable` -> `class UITable`

```
    return (
        <Table className="page--ui_table">
            ...
        </Table
    );
```

`class PageTableContainer` -> `class PageTable`

```
const isStickyColumns = stickyColumns > 0;
const containerClasses = ClassNames('ui', 'page--table', className, {
    page--table-sticky_columns': isStickyColumns,
});
const newStyle = Object.assign({}, style, {
    margin: bleed ? '0 -22px' : null,
});

if (isStickyColumns) {
    return (
        <div
            className={containerClasses}
            style={newStyle}
        >
           <div className="page--table_fixed_body">
               <PageTable
                   {...this.props}
                   collapsed={collapsed}
                   idPrefix="body"
                   onScrollColumn={this._onScrollColumn}
                   ref={ref => this._bodyTable = ref}
                   sizes={sizes}
                   style={{ minWidth }}
               />
           </div>
           
           <div className="page--table_fixed_column">
               <PageTable
                   {...this.props}
                   columns={_.slice(columns, 0, stickyColumns)}
                   idPrefix="column"
                   sizes={sizes}
                   onSplitter={this._onSplitterClick}
               />
           </div>
        </div>
    );
} else {
   return (
        <div
            className={containerClasses}
            style={newStyle}
        >
            <PageTable {...this.props} />
        </div>
    );
}
```

src/modules/pageTable.js
```diff
@@ -119,23 +175,201 @@ class PageTable extends React.PureComponent {
...
+    sizes: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.object)),
+    small: PropTypes.bool,
+    splitter: PropTypes.bool,
+    style: PropTypes.object,
+};
+
+class PageTableContainer extends React.Component {
+    constructor(props) {
+        super(props);
+        this._onResize = this._onResize.bind(this);
```
morethanfire: We need to put `this._onResize` in a debounce function and make a the handle draggable.

src/modules/pageTable.js
IlyaRadinsky: I wanted to implement scroll right for non sticked columns on click - I've tested again, it works badly. So, I've restored  `isClickable` as it was.

src/modules/pageTable.js
IlyaRadinsky: Fixed

src/modules/pageTable.js
IlyaRadinsky: Restored as it was.

src/modules/pageTable.js
IlyaRadinsky: Done.

src/modules/pageTable.js
```diff
@@ -2,6 +2,8 @@ import _ from 'lodash';
 import ClassNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
+import ReactDOM from 'react-dom';
+import SplitterSvg from './splitter.svg';
```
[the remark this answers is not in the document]
IlyaRadinsky: Done.

src/modules/pageTable.js
IlyaRadinsky: Done.

src/modules/pageTable.js
IlyaRadinsky: Fixed.

src/elements/icon.js
```diff
@@ -3607,6 +3607,41 @@ class Icon extends React.Component {
...
+                                                    {renderGradientColor ? (
+                                                        <rect fill={`url(#${gradientId})`} mask={`url(#${maskId})`} x="0" y="0" width="5px" height="14px" />
+                                                    ) : null}
+                                                </g>
+                                            </g>
+                                        </g>
+                                    </g>
+                                </g>
+                            </g>
+                        </g>
```
morethanfire: Is there a need for all these groups (`<g />`) elements? Also, I am a little concerned with the IDs as well with having the multiple icons of the same type on a page.

IlyaRadinsky: Fixed

IlyaRadinsky:
@morethanfire will do shadow and resizing fixed column by drag

IlyaRadinsky:
@morethanfire @groberts314 I've pushed the 'drop shadow' and the 'resize column' features, it works for both desktop and mobile. Here is a mini demo:

resizing_column

groberts314:
Looking _great_ to me!

src/modules/pageTable.js
```diff
@@ -119,23 +188,285 @@ class PageTable extends React.PureComponent {
...
+PageTableContainer.propTypes = {
     bleed: PropTypes.bool,
     className: PropTypes.string,
     columns: PropTypes.array.isRequired,
     data: PropTypes.array.isRequired,
     fontSize: PropTypes.string,
+    minWidth: PropTypes.number,
     rowProps: PropTypes.func,
     small: PropTypes.bool,
+    splitter: PropTypes.bool,
```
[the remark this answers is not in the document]
IlyaRadinsky: Fixed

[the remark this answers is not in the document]
IlyaRadinsky: Shadow fixed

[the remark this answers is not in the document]
IlyaRadinsky: About scrollbar: at the moment to implement sticked columns we have to render two tables: the first one - for sticked columns, the second - the full-width scrollable table. So, I'm pretty sure that to implement the column with own horizontal scrollbar we will have to render three tables: for the sticked columns, for the scrollable columns and the full-width table (we don't know about how the table columns will be rendered by browser, so, we need to render them first and then apply these widths for the sticked and for the scrollable columns).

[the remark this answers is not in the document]
IlyaRadinsky: Live column resizing is doable, will do.

IlyaRadinsky:
@morethanfire so, I've implemented the live column resizing, not sure that it works good (on desktop mode especially, mobile mode works better) - I'm watching the re-render performance problems:
live_resizing

live_resizing_mobile

As an option, I can try to make a live vertical line separator (thro the whole table height), but resize the columns only it on drag end.

IlyaRadinsky:
@morethanfire I've pushed the latest updates - so, the PR is ready to be merged and tested.

groberts314:
Thanks for merging @morethanfire 
Are we planning on doing a 6.3.0 that has this in it?

PR #59, opened 2019-09-23
[dev] [UI] AB#19144 Ability to view notes on a person record - [A.1]

fantaJinMode:
-- Added ability to make drawer component sticky and set `max-height` for those sticky drawers

Also updated CM UI docs,

Link  : modules/drawer/sticky


docs/src/modules/drawerSticky.js
```diff
@@ -0,0 +1,261 @@
+import { Button, Card, Drawer, Header, TitleBar } from 'react-cm-ui';
```
morethanfire: Thanks for doing documentation on this!

src/modules/drawer.js
```diff
@@ -387,6 +405,11 @@ Drawer.propTypes = {
     className: PropTypes.string,
     dimmer: PropTypes.bool,
     isOpen: PropTypes.bool.isRequired,
+    isSticky: PropTypes.bool,
+    maxHeight: PropTypes.oneOfType([
+        PropTypes.number,
+        PropTypes.string,
+    ]),
```
morethanfire: Let's just go with a number instead of both for `maxHeight`.

src/modules/drawer.scss
```diff
@@ -20,6 +20,9 @@ html:not(.ua-ie-edge):not(.ua-ie-11) body {
         width: 100%;
         z-index: $zindex + 2;
     }
+    &.drawer-sticky-mode{
+        position: unset !important; // sass-lint:disable-line no-important
```
morethanfire: Is this the rule that's causing the page to jump back to the top of the viewport when opening up the sticky drawer?

src/modules/drawer.scss
```diff
@@ -109,6 +112,21 @@ html:not(.ua-ie-edge):not(.ua-ie-11) body {
...
+        top: unset;
+        .drawer-container { 
+            bottom: 0;
+            top: unset;       
+            > div:first-child {
+                border-top-left-radius: 5px;
+            }
+        }
+        .shadow_container > div { 
+            box-shadow: -9px 0 17px 0 rgba(0, 0, 0, 0.1) !important;
```
morethanfire: Let's add the `sass-lint:disable-line no-important` comment on this line.

src/modules/drawer.js
```diff
@@ -387,6 +405,11 @@ Drawer.propTypes = {
     className: PropTypes.string,
     dimmer: PropTypes.bool,
     isOpen: PropTypes.bool.isRequired,
+    isSticky: PropTypes.bool,
```
morethanfire: Can we change this `isSticky` prop to `positionY` and have it be an enum. `[ 'bottom', 'top']`. The default string for this would be `top`.
I suppose this messes with your top left corner radius logic too. Sorry about that. I would say if `maxHeight` isn't 100% and `positionX` is `top` the bottom left corner would be rounded and visa-versa for the top left corner.

fantaJinMode: Fixed

src/modules/drawer.scss
fantaJinMode: Fixed

src/modules/drawer.scss
fantaJinMode: Yes, Because of it, Page is jumping back.

src/modules/drawer.js
fantaJinMode: Fixed

src/modules/drawer.scss
fantaJinMode: Fixed. It took me sometime to figure out problem and fixed it. Now, its completed.

docs/src/modules/drawerSubNavigation.js
```diff
@@ -114,9 +122,62 @@ class ModulesDrawerSubNavigation extends React.PureComponent {
                 <SubNavigation.Item label="Title Bar" />
                 <SubNavigation.Item label="Wing" />
                 <SubNavigation.Item label="Details" />
+                <SubNavigation.Item label="Sticky" />
             </SubNavigation>
         );
     }
+
+    _onSubNavClick(index) {
```
groberts314: Do we need this function?  It looks like a duplicate of `static onSubNavClick(index)`...?

src/modules/drawer.js
```diff
@@ -310,9 +318,12 @@ class Drawer extends React.Component {
 
                         break;
                 }
-
                 BODY.style.top = `-${scrollPosition}px`;
+
                 DOMUtils.addClassName(BODY, 'drawer-open');
+                if (positionY && maxHeight) {
+                    BODY.style.position = 'inherit';
+                }
```
groberts314: Nitpick; `if(...) { ... }` block should have an empty line before and after.

fantaJinMode: @groberts314 thanks, will correct it on my next PR on react-cm-ui

groberts314: Sounds good.  If it's not going to distract _too_ much or take _too_ long, maybe it's worth fixing all the ESLINT issues in that file at some point.

@morethanfire can weigh in, but in my own personal opinion, I'd _love_ to get the React CM UI codebase to a point where it is certified LINT free and we can run the linters (ES and Sass) as a part of builds, and fail builds if there are linter issues.  We probably will have an easier time of achieving that first that in React CM UI, _long_ before we would get there in HC (`church-management` repo `/client`), where our ESLINT related pile of tech debt is _ginormous!_

Side note: I had a thought about HC client application and ESLINT.  Maybe we can do something like the following:
* Run ESLINT in a mode where it will not fail the build, but will produce an artifact giving us statistics and/or a comprehensive list of the issues (i.e. redirect the output to a file which gets published as an artifact, and ensure the return code of the command is 0).
* "Baseline" it.  Let's say that, as of some point, we have ~ 10,000 errors and ~ 12,000 warnings (I'm making up numbers as an example).  What we can do as try to get it to fail the build only if the number of errors or warnings in the present run exceeds the baseline threshold values (these are build parameters/variables); i.e. it's a failure only if the changeset included in the new build has introduced new problems.  Then we periodically reduce the threshold values as we reduce the volume of legacy lint, slowly improving overall code quality.

Thoughts?

PR #60, opened 2019-09-25
[hotfix/6.1.1 -> dev] (HC [release/1.19.0]) Bug AB#21242 [Serving Opportunity] - Issue with ranged dates on Recurring tab under When & Where in the New Serving Opportunity drawer

groberts314:
* This PR fixes a bug in `<DatePickerInput />` where, if using two of these as 'From' and 'To' dates describing a range, when typing out the desired date in one of them (rather than using the calendar), it would actually clear both 'from' and 'to' dates (which was causing Bug [AB#21242](https://dev.azure.com/saddlebackchurch/b8a0d0bb-a4fe-48e3-9a60-2f1440145fc4/_workitems/edit/21242) in the HC back-office/admin application).
* Also cleaned up ESLINT issues in the documentation both for `<DatePickerInput />` and for `<Prompt />`


groberts314:
Note: we _cannot_ resolve the conflicts until we tag `6.1.1` and publish it to NPM.

PR #66, opened 2019-10-09
PBI AB#20349 [My Ministry > Members] [mobile] - Implement page details banner

IlyaRadinsky:
In this demo the `Expand` button showing in mobile mode only.

page-details-demo

@groberts314 , @morethanfire please have a look

### DIFF
--- a/docs/src/modules/pageDemo.js
+++ b/docs/src/modules/pageDemo.js
@@ -277,11 +277,11 @@ class PageDemo extends React.PureComponent {
                 fontWeight: 'bold',
                 header: null,
             }, {
-                accessor: 'activeTemplates',
-                header: 'Active Templates',
+                accessor: 'notContacted',
+                header: 'Not Contacted',
             }, {
-                accessor: 'inactiveTemplates',
-                header: 'Inactive Templates',
+                accessor: 'contacted',
+                header: 'Contacted',
             },
         ];
 
@@ -289,11 +289,14 @@ class PageDemo extends React.PureComponent {
             {
                 columns: [
                     {
-                        accessor: 'activeTemplates',
-                        header: 'Active Templates',
+                        accessor: 'firstContact',
+                        header: '1st Contact',
                     }, {
-                        accessor: 'inactiveTemplates',
-                        header: 'Inactive Templates',
+                        accessor: 'secondContact',
+                        header: '2nd Contact',
+                    }, {
+                        accessor: 'pending',
+                        header: 'Pending',
                     },
                 ],
                 divide: true,
@@ -490,9 +493,11 @@ class PageDemo extends React.PureComponent {
                                 color={11}
                                 columns={bannerColumns}
                                 data={{
-                                    activeTemplates: 4,
-                                    id: 1,
-                                    inactiveTemplates: 2,
+                                    notContacted: 4,
+                                    contacted: 5,
+                                    firstContact: 4,
+                                    secondContact: 1,
+                                    pending: 3,
                                 }}
                                 detailedColumns={bannerDetailedColumns}
                             />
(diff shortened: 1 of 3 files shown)


groberts314:
Is there any way to align the second row of statistics with the start of the first (leaving the space below the chart empty), like the mocks for My Ministry?

Members Page Details Interested Expanded

Members Page Details Mobile Expanded

docs/src/modules/pageDemo.js
```diff
@@ -262,6 +262,47 @@ class PageDemo extends React.PureComponent {
...
+                        accessor: 'activeTemplates',
+                        header: 'Active Templates',
+                    }, {
+                        accessor: 'inactiveTemplates',
+                        header: 'Inactive Templates',
+                    },
+                ],
+                divide: true,
+            },
+        ];
```
groberts314: For demo purposes could we give the extra columns different names and different counts.  e.g. "Foo Templates" and "Bar Templates" maybe?  So we can tell that potentially different data is in the collapsible area?

IlyaRadinsky:
@groberts314 yes, I saw the alignment problem, and we should decide how to resolve it (if it possible). The problem is that actually in demo page banner we have 3 columns:
- tittle (full width)
- stats column
- extra stats column

So, to align the `Expand` button with `stats column` there are should be in the same row and there are should be ability to pull right the `Expand` button. In my opinion, as the result, the config for columns will be tricky and complicated. So, after some thought, I've decided to present this variant - when all title/stats columns are in a single (and wrapable) column, and `Expand` button - is another one pulled right.

As another option, we could add `expandButtonMarginTop` prop for Page.Details.

IlyaRadinsky:
page-details-demo2

groberts314:
> So, after some thought, I've decided to present this variant - when all title/stats columns are in a single (and wrapable) column, and `Expand` button - is another one pulled right.

If I understand this option, correctly, this sounds like a good plan.  Will you be able to update this by tomorrow?  Did you need any further feedback from @morethanfire or me?

IlyaRadinsky:
@groberts314 this is already done

groberts314:
Oh, sorry, I thought it was still TODO.  Mainly because the screenshot still shows the second row left-aligned under the "chart" placeholder.  Do you have a screen shot of it with the desired alignment?

IlyaRadinsky:
@groberts314 

Screen Shot 2019-10-09 at 7 40 30 PM

groberts314:
image

IlyaRadinsky:
Ah, I see... 
So, the alignment of `Not Contacted` with `1st Contact` we can get by adding the invisible stub for `extra stats column` (with the same size as the donut chart). The question is about margin top for `Expand` button. If it OK to add `expandButtonMarginTop` to Page.Details - I'll do

groberts314:
> So, the alignment of Not Contacted with 1st Contact we can get by adding the invisible stub for extra stats column (with the same size as the donut chart).

Okay, got it, so this should be dealt with on consumer side when configuring the columns.  Makes sense.

> The question is about margin top for Expand button. If it OK to add `expandButtonMarginTop` to `Page.Details` - I'll do

@morethanfire How do you feel about `expandButtonMarginTop` prop?

groberts314:
@IlyaRadinsky I'm hoping we're going to :shipit: on this one today shortly here, and release it in React CM UI `6.4.1`.  Assuming that's the case, will you be able to complete the HC side and submit a PR tomorrow?

IlyaRadinsky:
@groberts314 yes, will do

src/modules/details.js
```diff
@@ -147,6 +165,21 @@ class Details extends React.PureComponent {
                             );
                         })}
                     </div>
+                    {hasDetailedColumns && (
+                        <Button
+                            color="light"
+                            icon
```
groberts314: We should probably add a `className` and/or `id` here.

PR #68, opened 2019-10-09
Fixed Details Eslint Issues

morethanfire:


src/modules/details.js
```diff
@@ -1,123 +1,227 @@
...
-        if (_.isString(column.accessor)) {
-            accessor = _.get(data, column.accessor);
-        } else if (_.isFunction(column.accessor)) {
-            accessor = column.accessor(data);
-        }
+    if (_.isString(accessor)) {
+        newAccessor = _.get(data, accessor);
+    } else if (_.isFunction(accessor)) {
+        newAccessor = accessor(data);
+    }
```
groberts314: `newAccessor` should be `accessedData` maybe?  i.e. it's the value we've extracted from `data` using the `accessor`.

PR #76, opened 2019-10-19
Added Time From Now Function To DateUtils

morethanfire:


src/utils/dateUtils.js
```diff
@@ -223,6 +223,46 @@ const DateUtils = {
...
+                h,
+                hh,
+                d,
+                dd,
+                M,
+                MM,
+                y,
+                yy,
+            },
+        });
```
groberts314: I imagine this actually mutates some state inside the global `moment` function/object definition?  Are we okay with that?  We've probably been doing it all along, so it's probably okay.

morethanfire: It does, but using Moment's `fromNow` function I am not sure how else to go about it. Do you have any ideas?

groberts314: Nah, not really.  It seems we essentially want to customize this part of moment's locale settings.  In the future, we might consider some research into how moment handles these locale settings and maybe find some way to set these locale settings once, and not _every time_ our `timeFromNow()` function is called.  But that doesn't have to block this PR.

PR #78, opened 2019-10-24
iOS Drawer header fix and renaming of dateUtils and domUtils

morethanfire:


src/index.js
```diff
@@ -39,8 +39,8 @@ export { default as TimePicker } from './modules/timePicker.js';
 
 // utils
 export { default as DatePickerUtils } from './utils/datePickerUtils.js';
-export { default as DateUtils } from './utils/dateUtils.js';
-export { default as DOMUtils } from './utils/domUtils.js';
+export { default as dateUtils } from './utils/dateUtils.js';
```
groberts314: Gonna be a find/replace for the imports in HC I imagine?

src/modules/drawer.js
```diff
@@ -369,20 +299,85 @@ class Drawer extends React.Component {
...
+                <div
+                    className={drawerClasses}
+                    ref={(ref) => { this.drawerRef = ref; }}
+                >
+                    <div
+                        className="drawer-container"
+                        ref={(ref) => { this.drawerContainerRef = ref; }}
+                        style={{
+                            height: _.isNumber(positionYOffset) ? `calc(100% - ${positionYOffset}px)` : null,
+                            // transform: this.setStartOfTransform(),
```
groberts314: Is there a good reason to keep this commented out?

src/index.js
morethanfire: Yeah. I thought about not doing this change just yet, but I don't believe it'll be that time consuming go forth.

src/modules/drawer.js
morethanfire: removed.

groberts314:
:shipit:

PR #89, opened 2019-11-19
AB#21616 - Data Groups Fix

morethanfire:


docs/src/modules/pageDemo.js
```diff
@@ -1041,55 +1065,58 @@ class PageDemo extends React.PureComponent {
...
-                            />
+                            >
+                                <Page.DataGroups
+                                    columns={mainCoulumn}
+                                    data={mainData}
+                                />
+                            </div>
                         </Page.Content>
                     </Page.Container>
                 </Page>
```
groberts314: Not blocking, but I'm sensing that the Demo Page is morphing from a generic "list of things" page to be more like the person 2.0 page.  Do we perhaps need multiple independent Demo Pages for different use cases?

src/elements/activityIndicator.js
```diff
@@ -181,10 +181,11 @@ class ActivityIndicator extends React.PureComponent {
             <div
                 className={containerClasses}
                 id={id}
-                style={({
+                style={{
                     ...style,
-                    transform: `scale(${(1 / 68) * size}) ${style && style.transform}`,
-                })}
+                    transform: `scale(${(1 / 68) * size})`,
```
groberts314: Is 1/68 some kind of "magic number"?  Would it be worth declaring a `const` to describe what this value represents?  Not blocking.

src/modules/dataGroups.js
```diff
@@ -37,40 +35,47 @@ class DataGroups extends React.PureComponent {
...
     }
 
-    reorderColumns(columnCount) {
+    reorderColumns() {
         const { columns } = this.props;
-        const newColumns = [];
+        const blockSize = this.dataGroups.offsetWidth;
+        const columnMinWidth = 300;
+        const columnsLength = columns.length;
+        let amountOfColumns = Math.floor(blockSize / columnMinWidth);
```
groberts314: `numberOfColumns` reads slightly better here (i.e. is more grammatically correct) than `amountOfColoumns`.  The difference between "number of" and "amount of" is subtle, having to do with whether the quantity being described is easily enumerable (countable), or if it is really, really large; see https://writingexplained.org/amount-vs-number-difference

src/modules/dataGroups.js
```diff
@@ -37,40 +35,47 @@ class DataGroups extends React.PureComponent {
...
         let columnNum = 0;
 
-        while (columnNum < columnCount) {
-            for (let i = 0; i < columns.length; i += columnCount) {
+        while (columnNum < amountOfColumns) {
+            for (let i = 0; i < columns.length; i += amountOfColumns) {
                 const column = columns[i + columnNum];
 
                 if (column !== undefined) {
-                    newColumns.push(column);
```
groberts314: `push()` is not wrong here per se, since `newColumns` is a local variable, so we don't have to avoid mutating it.  It's fine the way you have it as well, but just letting you know.  AFAIK, immutability is mostly important for shared values, i.e. from component state, Redux state or Flux state.  But it is always proper to mutate a local variable within the scope of a function body.

src/elements/activityIndicator.js
morethanfire: Fixed.

docs/src/modules/pageDemo.js
morethanfire: I think this is a good idea, I will try to keep it in mind when have more time to give.

src/modules/dataGroups.js
morethanfire: Fixed.

docs/src/modules/pageDemo.js
groberts314: Yep, fair enough.

PR #90, opened 2019-11-22
DataGrid Bug Fix

morethanfire:


src/modules/dataGrid.js
```diff
@@ -69,7 +69,7 @@ class DataGrid extends React.Component {
             const row = [];
 
             for (let j = 0; j < stickyColumns; j += 1) {
-                const el = document.querySelector(`#table--table_cell-body-${i}-${j}`);
+                const el = document.querySelector(`#page--data_grid_table_cell_column-${i}_${j}`);
```
groberts314: Does `page` need to be parameterized?  Is this one of those things you use in a Drawer as well as a Page?

PR #91, opened 2019-11-22
Feature/release 7.2.2  data grid handle bug

morethanfire:


morethanfire:
I tested these changes in the react-cm-ui page demo and on HC in the My Ministry roster pages.

src/modules/dataGrid.scss
```diff
@@ -26,18 +26,25 @@ $blackAlpha: rgba(0, 0, 0, .5);
         }
         &_table_component {
             .table-cell:last-child {
-                padding-right: inherit;
+                padding-right: 11px !important;
```
groberts314: I imagine `!important` is in fact important? ;-)

groberts314: Do we need a SASS Lint comment?

src/modules/dataGrid.scss
```diff
@@ -26,18 +26,25 @@ $blackAlpha: rgba(0, 0, 0, .5);
...
-                    float: right;
-                    position: relative;
-                    .ui.icon-size-small {
-                        width: 5px;
-                        svg {
+                padding-right: 11px !important;
+                position: relative;
+                .#{$prefixDrawerDataGrid}, .#{$prefixPageDataGrid}  {
+                    &_table_header_handle {
+                        cursor: pointer;
```
groberts314: Should we consider `cursor: col-resize;`?

morethanfire: Yep, i like it.

src/modules/dataGrid.scss
morethanfire: Added.

morethanfire: Yeah, they need to be important because of the old school way of class names taking priority.

groberts314: Stinkin' legacy CSS.  Got it. 😁

PR #93, opened 2019-12-17
[dev] AB#22227 - Filters Drawer and Filters Rail For Page and Drawer, Checkbox and Toggle added to Filters Drawer

morethanfire:
Features and Fixes in this PR:

**New `./src/` File Structure Beginnings**
- Moved all template files to `./src/templates`.
- `./src/templates` JS/JSX files are Eslint free.
- Moved `./docs/` package.json `dependencies` and `devDependencies` to `./docs/` new package.json
- Moved webpack.config to `./docs/`
- implemented .bablerc settings into new babel.config.js file.
- Deleted .babelrc

**FiltersDrawer:**
- Copied old pageFiltersDrawer.js and created filtersDrawer.jsx. Also split its components out into individual files.
- Copied old pageFiltersDrawer.scss and created filtersDrawer.scss
- Renamed old pageFiltersDrawer.js -> pageDeprectatedFiltersDrawer.jsx
- Renamed old pageFiltersDrawer.scss -> pageDeprecatedFiltersDrawer.scss
- The components pageFiltersDrawer.jsx and drawerFiltersDrawer.jsx wrap `<FiltersDrawer />` now.
- Added `checkbox` and `toggle` to <FiltersDrawer />
- Copied/Renamed `./docs/modules/drawerFiltersDrawer.js` -> `./docs/tempaltes/drawerFiltersDrawer.jsx`.
- Created a new Drawer Deprecated Filters Drawer doc page, `./docs/modules/drawerDeprecatedFiltersDrawer.js`

**Container**
- Copied pageContainer.js and created container.jsx
- Changed pageContainer.js -> pageContainer.jsx and wrapped container.jsx
- Created a new drawer.Container.jsx

**Content**
- Created content.jsx
- Changed pageContent.js -> pageContent.jsx and wrapped content.jsx
- Changed drawerContent.js -> drawerContent.jsx and wrapped content.jsx
- Copied drawerContent.scss and implemented into content.scss
- Copied pageContent.scss and implemented into content.scss
- Created a new drawer.Container.jsx


src/templates/content.jsx
```diff
@@ -0,0 +1,57 @@
...
+    as: 'section',
+    children: undefined,
+    className: undefined,
+    id: undefined,
+    isFiltersRailOpen: false,
+    scrollable: false,
+    style: {},
+};
+
+function PageContent(props) {
```
groberts314: Should this simply be `Content` rather than `PageContent`?

src/templates/dataGridTable.jsx
```diff
@@ -3,10 +3,43 @@ import ClassNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import DataGridTableRow from './dataGridTableRow.js';
-import DragListener from '../utils/dragListener.js';
+import DataGridTableRow from './dataGridTableRow';
```
groberts314: We are deliberately going with no file extensions now?  When did that change?

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+
+            if (key === 'value') {
+                newData = selectedOptions;
+            }
+
+            return newData;
+        });
+
+        this.setState({
+            nestedTogglesData: _.cloneDeep(newNestedTogglesData),
```
groberts314: I don't think `_.cloneDeep()` is necessary here.  `newNestedTogglesData` is already a brand spanking new data structure that you've freshly instantiated in here; I think it is okay if it directly is assigned to `nestedTogglesData` in the `setState()` call.

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+            id,
+            isDirty,
+            isFiltering,
+            isOpen,
+            onClose,
+            rows,
+            style,
+        } = this.props;
+        const { nestedTogglesData } = this.state;
+        const containerClasses = ClassNames('ui', 'page--filters_drawer', className);
```
groberts314: Do we need our now classic parameterization of `page` versus `drawer` here?

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+                    <Drawer.TitleBar
+                        className="page_filters_drawer--title_bar"
+                        closeButton={(
+                            <Icon
+                                className="page_filters_drawer--close_button"
+                                compact
+                                onClick={onClose}
+                                type="times"
+                            />
+                    )}
```
groberts314: Whitespace is off here, yeah?  ESLINT didn't get on your case about it?

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+                                >
+                                    {/* eslint-disable max-len */}
+                                    {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions, jsx-a11y/anchor-is-valid */}
+                                    {canClear ? (
+                                        <a
+                                            className={clearFiltersClasses}
+                                            onClick={this.onClearClick}
+                                        >
+                                            Clear Filters
+                                        </a>
```
groberts314: I'm wondering if we need to be able to start styling buttons to look like links instead of using `<a>` tags and always having to suppress all those accessibility warnings.  Clearly not the hottest priority, but it does seem like we're "doing it wrong", perhaps.

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+                                    </Button>
+                                    )}
+                                </div>
+                            </div>
+                        )}
+                    />
+
+                    <Drawer.Content>
+                        {_.map(rows, (row) => {
+                            rowKeyNum += 1;
```
groberts314: How is this any different from simply using the array index as a key? :-)
Not actionable; I'm sure it's just fine as-is.  I'm just curious about the thinking here.

src/templates/filtersDrawer.jsx
```diff
@@ -0,0 +1,674 @@
...
+                                                nestedToggles,
+                                                toggle,
+                                            } = item;
+                                            const itemClassName = ClassNames('page_filters_drawer--item', {
+                                                'page_filters_drawer--item-checkbox': checkbox,
+                                                'page_filters_drawer--item-dropdown': dropdown,
+                                                'page_filters_drawer--item-jsx': jsx,
+                                                'page_filters_drawer--item-multi_select': multiSelect,
+                                                'page_filters_drawer--item-nested_toggles': nestedToggles,
+                                                'page_filters_drawer--item-toggle': toggle,
```
groberts314: Do we need to worry about `page` versus `drawer` here?

src/templates/filtersDrawerNestedTogglesLabel.jsx
```diff
@@ -0,0 +1,59 @@
...
+    onClick() {
+        const { nestedTogglesData, onClick } = this.props;
+
+        onClick(nestedTogglesData);
+    }
+
+    onKeyDown(event) {
+        const { nestedTogglesData, onClick } = this.props;
+
+        if (event.keyCode === 13) {
```
groberts314: Don't we have a `const` for this somewhere?

src/templates/page.jsx
```diff
@@ -80,26 +96,15 @@ Page.Content = PageContent;
...
-Page.Details = PageDetailsWindow; // TODO: Deprecated. Alias name for Page.DetailsWindow. Remove in next major release.
+Page.DeprecatedFiltersDrawer = PageDeprecatedFiltersDrawer; // TODO: Deprecated. Remove in a major release.
+Page.Details = PageDetailsWindow; // TODO: Deprecated. Alias name for Page.DetailsWindow. Remove in a major release.
 Page.DetailsWindow = PageDetailsWindow;
-Page.FiltersDrawer = PageFiltersDrawer;
+Page.FiltersDrawer = PageFiltersDrawer; // TODO: Deprecated. Remove in a major release.
 Page.FiltersRail = PageFiltersRail;
-Page.Grid = PageDataCards; // TODO: Deprecated. Alias name for Page.DataCard. Remove in next major release.
-Page.Table = PageDataGrid; // TODO: Deprecated. Alias name for Page.DataGrid. Remove in next major release.
+Page.Grid = PageDataCards; // TODO: Deprecated. Alias name for Page.DataCard. Remove in a major release.
```
groberts314: Should it be `Page.DataCard` (singular) or `Page.DataCards` (plural)?

src/templates/pageDeprecatedFiltersDrawer.jsx
```diff
@@ -170,14 +172,83 @@ class PageFiltersDrawer extends React.Component {
...
+
+        if (_.some(nestedTogglesData.value, selectedOption)) { // Subtract
+            selectedOptions = _.filter(nestedTogglesData.value, d => d.value !== selectedOption.value);
+        } else { // Add
+            selectedOptions = _.sortBy([ ...nestedTogglesData.value, selectedOption ], [ 'value' ]);
+        }
+
+        const newNestedTogglesData = _.mapValues(nestedTogglesData, (d, key) => {
+            if (key === 'value') {
+                d = selectedOptions;
```
groberts314: We're intentionally assigning to (mutating) an iteration variable here?  That's unusual.  But it is the correct logic and is being done intentionally?

src/modules/drawerDeprecated.js
```diff
@@ -1,11 +1,11 @@
-'use strict';
+/* eslint-disable */
```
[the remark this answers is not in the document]
morethanfire: Yeah, that's the thought.

src/templates/content.jsx
morethanfire: Yes, it should. Fixing.

src/templates/dataGridTable.jsx
morethanfire: Yes, but only for react-cm-ui. The decision was made because of switching to `.jsx` extensions and Babel isn't taking a `.jsx` file and spitting out babelfied file with the same extension. In other words, it's an easy fix.

src/templates/filtersDrawer.jsx
morethanfire: Ding Dang! Yep we do.

src/templates/drawerContent.jsx
```diff
@@ -0,0 +1,54 @@
...
+    as: 'section',
+    children: undefined,
+    className: undefined,
+    id: undefined,
+    isFiltersRailOpen: false,
+    scrollable: false,
+    style: {},
+};
+
+function PageContent(props) {
```
fantaJinMode: I think there should be `DrawerContent`

src/templates/filtersDrawer.jsx
morethanfire: Fixed.

src/templates/filtersDrawer.jsx
morethanfire: I don't believe we're "doing it wrong" necessarily, but I do agree we need to figure out a different approach. I was actually thinking about the `<Link />` react-router component, https://knowbody.github.io/react-router-docs/api/Link.html. We'd need to look more into the component.

src/templates/filtersDrawer.jsx
morethanfire: My thinking here is: don't ever use the index even if i know the list isn't going to change its sort order. Use an ID whenever possible and when not, use a global index.

src/templates/filtersDrawer.jsx
morethanfire: Fixed.

src/templates/filtersDrawerNestedTogglesLabel.jsx
morethanfire: Not yet. Do we need a `const`?

src/templates/page.jsx
morethanfire: `Page/Drawer.DataCards`. Plural because there can be more than one. It's my belief that `Page/Drawer.DataGrid` is plural as well, because "grid" can be referring to one or more.

src/templates/pageDeprecatedFiltersDrawer.jsx
morethanfire: It was intentional, not saying it was the best though.

src/templates/filtersDrawerNestedTogglesLabel.jsx
groberts314: I see this littered all over HC:
```JavaScript
const ENTER_KEY_CODE = 13;

class FooComponent extends React.PureComponent {

// ... blah blah blah ...

    onKeyDown(event) {
        if (event.keyCode === ENTER_KEY_CODE) {
            // do whatever it is supposed to do when 'Enter' was hit...
        }
    }

    // ... more blah blah blah ...
}
```

At some point we had discussed moving `const ENTER_KEY_CODE = 13;` into some shared constants file, exporting it from there and importing it everywhere it is needed, but apparently we never pulled the trigger on that.  Is it worth doing something like that in both projects (React-CM-UI and HC), or is that overkill?

src/templates/page.jsx
groberts314: Is the actual alias correct then and just the comment is off?

src/templates/dataGridTable.jsx
groberts314: Understood (well, understood-ish) ... it solved a problem. 👍

src/templates/filtersDrawer.jsx
groberts314: Okay, so, in your opinion, this is preferred in cases where there is no ID or other "natural" key to help reinforce that one should _never_ use the array key, period?   I'll buy that.

src/templates/filtersDrawer.jsx
groberts314: Wouldn't that just be masking the problem? :-)

groberts314: I guess I mean "doing it wrong" from an accessibility standpoint: in other words, are we using hyperlinks in a way that's not intended, and this has accessibility implications.  That's been my understanding of the warnings.  Maybe I'm reading too much into it.  We've not made accessibility a priority at all, so suppressing the warnings is the right approach in the short-term ... I guess I'm wondering what we would do if we were to actually heed the warnings?

src/templates/page.jsx
morethanfire: The comment is off. Sorry. :(

src/templates/filtersDrawer.jsx
morethanfire: Yep, that's my personal take. Doesn't have to be anyone else's though.

groberts314: I understand.  It does have the benefit of allowing all developers to look at the same code and _always_ know that "array indexes as keys are an anti-pattern".  That's an easier policy to grok than "sometimes using array indexes as keys are okay and other times it's bad; it depends!"

So ... I'm buying what you're selling.

src/templates/filtersDrawerNestedTogglesLabel.jsx
morethanfire: Yeah, you're right, I think it would be better to have a util `function` to handle the problem instead of a util `const`.  What do you think?

groberts314: something like this?
`utils/keyEventUtils.js`
```JavaScript
const ENTER_KEY_CODE = 13;

// ...

export function isEnterKeyEvent(event) {
    return _.isObject(event) && event.keyCode === ENTER_KEY_CODE;
}
```

PR #94, opened 2019-12-18
Feature/dropdown year bug fix

Benjamin-Gnewuch:
Fix for [PBI#23545](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/23545)


src/modules/datePickerDropdownYear.js
```diff
@@ -74,6 +74,7 @@ class DatePickerDropdownYear extends React.PureComponent {
                 <ScrollBar
                     autoHide
                     className="date-picker-dropdown-year-scrollbar"
+                    ref={ref => this.scrollComponent = ref}
```
groberts314: I think ESLINT is demanding this these days:
`ref={(ref) => { this.scrollComponent = ref; }}`

Benjamin-Gnewuch:
@groberts314 
The complaint about our year dropdown select is that even though the value defaults to 2019, when the user expands the dropdown to change to a different year, it starts at the top of the list, which is 100 years into the future.

image

This change makes it scroll to the currently selected value on expanding.

groberts314:
Thanks @Benjamin-Gnewuch ... that helps.  And that's clearly suboptimal UX to say the least! ;-)
I'm gonna keep this open until we get a review from one of the UI experts.

src/modules/datePickerDropdownYear.js
```diff
@@ -87,11 +88,10 @@ class DatePickerDropdownYear extends React.PureComponent {
         const { maxDate } = this.props;
 
         if (!maxDate) {
-            const datePickerWrapElement = ReactDOM.findDOMNode(document.querySelector('.date-picker-dropdown-year-scrollbar > div'));
             const selectedElement = ReactDOM.findDOMNode(document.querySelector('.date-picker-year-option-is-selected'));
             const selectedPosistion = selectedElement.offsetTop - 11;
 
-            domUtils.scrollTo(selectedPosistion, 0, datePickerWrapElement);
+            this.scrollComponent.scrollTop(selectedPosistion);
```
morethanfire: Love the ref being used here and removing the `domUtils.scrollTo()` function. It's not really all that needed to see an scrolling animation.

PR #95, opened 2019-12-24
[dev] Table Row Reordering with Drag and Drop

groberts314:
This feature is being added to React CM UI's `<Table>` (specifically `<Table.Row>`) in order to power **[Product backlog Item AB#22125](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/22125) | Ability to manage library of recommended ministries - [K.1.2]**

Also, I refactored the Documentation page for the `<Table>` component to divide it into many smaller child components (one for each sample or related group of samples) and to otherwise address ESLINT gripes (e.g. JS => JSX, etc.)

image

image

I'm not sure how to produce a video/animated GIF ... sorry :-(


IlyaRadinsky:
@groberts314 just now I've tested this branch in mobile mode and it doesn't work to me :(

groberts314:
@IlyaRadinsky Thanks for the feedback and testing; I greatly appreciate it.  Do you have any suggestions on extending this for mobile touch?  That's a bit outside of my normal areas of expertise, so I could probably use some assistance from yourself or Mssr. @morethanfire .  I'm curious ... does the custom fields question choice reordering you built in HC work for mobile?  I tried to follow that as a guide for this.  If so, I wonder if it is something having to do with trying to use it in a table specifically.

I imagine that, in terms of the Recommended Ministries feature in HC which is first up to leverage this, no mobile is probably not a showstopper; we could do our thing to put the messaging that the page is only available in desktop mode.  We'll of course have to clear this with the business if we cannot make this work on mobile, but it seems reasonable to me to ship a first cut of the feature. 
 Longer term though, I'd imagine it would be good to get this to work on mobile, especially if the designers want to use this pattern for other features.

Just a reminder all that I am out of the country and not working from 26 Dec 2019 - 7 Jan 2010.  I'll resume working on 8 Jan.

IlyaRadinsky:
@groberts314 look at `onTouch*` events, I used them for `Page.DataGrid` component.

docs/package.json
```diff
@@ -3,12 +3,16 @@
...
+    "build:clean": "rimraf build",
     "start": "webpack-dev-server --mode development --content-base docs/build/ --hot --progress --colors --host 0.0.0.0",
     "start:stats": "webpack --display-modules --display-chunks --json > stats.json --progress"
   },
   "dependencies": {
     "@babel/register": "^7.4.4",
     "core-js": "^3.1.3",
+    "lodash": "^4.17.14",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
```
morethanfire: NAB: A little curious to know if there is any issues that come along with adding react here. Like, when running the docs build if there is any errors yelling at us about having two react modules installed.

groberts314: @morethanfire I hadn't noticed any issues with adding it.  And ESLINT was complaining to me about not having it.  I assume since merging this things are all still happy.

@morethanfire @IlyaRadinsky I guess I can try and handle the `onTouch*` events for mobile/touch support in a subsequent PR?  Is the browser based emulator typically good enough for testing this, or is NGrok and actual device testing recommended?

groberts314:
@morethanfire Thanks for merging this!  It looks like it is available as part of React CM UI 8.0.0.

@morethanfire @IlyaRadinsky I guess I can try and handle the `onTouch*` events for mobile/touch support in a subsequent PR?  Is the browser based emulator typically good enough for testing this, or is NGrok and actual device testing recommended?

PR #97, opened 2019-12-25
[dev] AB#22234 Ability for a serving opportunity administrator to check in/out a volunteer [B.1] - Room Mgmt [OLD]

IlyaRadinsky:
Added the ability to specify an own ID for `DataGrid` - each cell of the `DataGrid` has an own ID, this fix is needed to be able to have multiple instances of the `DataGrid` on one page.


IlyaRadinsky:
@groberts314 @morethanfire please have a look.

PR #98, opened 2019-12-25
[dev] AB#22234 Ability for a serving opportunity administrator to check in/out a volunteer [B.1] - Room Mgmt [OLD]

IlyaRadinsky:
Fixes getting `DataGrid` row height on column resize


IlyaRadinsky:
PR #97 should be merged first.

PR #99, opened 2020-01-07
Fixed typo 'Decemeber' in month picker

Benjamin-Gnewuch:


Benjamin-Gnewuch:
@morethanfire Cris just pointed out this is a thing, so I made a quick fix for it. I'm not aware of any official report of this typo.

PR #102, opened 2020-01-09
Proposed 9.0.0 Release 

morethanfire:
#### Feature
- Commit 59b9716
    - Added `className` and `id` to **HeaderSubHeader**.
- Commit eca4611
    - Added `form` to the **Grid's** propTypes. This will give a **Grid** the vertical negative space it needs between each **Grid.Column** housing a form field.
- Commit 0daf4e9
    - Renamed the dateUtils "format" functions for better use.

#### Bug Fix
- Commit 0daf4e9
    - Fixes the list in **ActionBar** from being centered.


src/elements/headerSubheader.jsx
```diff
@@ -0,0 +1,41 @@
...
+    children: PropTypes.node.isRequired,
+    className: PropTypes.string,
+    id: PropTypes.string,
+    style: PropTypes.shape({}),
+};
+
+const defaultProps = {
+    className: undefined,
+    id: undefined,
+    style: {},
```
groberts314: Is it useful for `style` prop to default to an empty object versus `null` or `undefined`?  This is probably not a super important concern; I'm mostly just curious. ;-)

src/templates/actionBar.scss
```diff
@@ -98,7 +98,6 @@ $backwards-step4-duration: $standard-fifth-timing;
     &.action_bar--list {
         align-items: center;
         display: flex;
-        justify-content: center;
```
groberts314: Any potential for style regressions here?

src/utils/dateUtils.js
```diff
@@ -94,40 +76,58 @@ const dateUtils = {
...
+    formatTime(data) {
+        if (_.isNil(data)) {
+            return null;
+        }
+
+        if (_.isNumber(data)) {
+            return moment.unix(data).utc().format('LT');
+        }
+
+        return moment(data, 'HH:mm').format('hh:mm a');
```
groberts314: I realize the original code was like this, but shouldn't this case also use `.format('LT')`?

src/templates/actionBar.scss
morethanfire: I don’t believe I check in a hand of locations that have a tid-bit different set of columns and they are good.

groberts314: Sounds good.  All you can do.  Just thought I'd ask it during a code review. 👍

src/elements/headerSubheader.jsx
morethanfire: You know, I have been thinking the same thing. I have going with empty object, because that’s what I started doing in the beginning of implementing the Airbnb Eslint rules. I think I would rather have undefined.

src/utils/dateUtils.js
morethanfire: Yep, but I though maybe there would be to much a regression?

groberts314: I reviewed the documentation (see https://momentjs.com/docs/#/displaying/format/) and  I believe the only difference (at least for the en-US locale) would be whether "AM/PM" is capitalized (as it is with the "LT" standard format) or lower-cased (with the "a" component format).

My own opinion is that we should try and display values in consistent formats, and therefore we should leverage the built-in standard ones, versus custom ones we build from the individual portions of a Date/Time value helps with this.  Also, with internationalization/localization in mind, it is arguably _"wrong"_ to "hard-code" format patterns such as "MM/DD/YY" since certain other locales besides en-US may favor variations such "DD/MM/YY" instead.  The built in standard formats (e.g. "L" for dates) will respect locale settings and do the right thing (see https://momentjs.com/docs/#/i18n/).

Furthermore, in my opinion, there is no logical reason that different forms of the input value (in this case a number representing a UNIX timestamp versus a string representing HH:mm) should result in differently formatted display values.  I cannot think of a valid justification for that behavior, at least off the top of my head.  In all likelihood the original implementation there that does result in this difference was simply an accident.

HOWEVER, all that having been said, your concern is noted, and if your preference for this time is to let it be and not change anything unnecessarily, I fully support that decision.

morethanfire: > My own opinion is that we should try and display values in consistent formats, and therefore we should leverage the built-in standard ones

I agree, this is why i renamed them in order to say, "hey, don't use that format like that use our dateUtils format functions."

I am going to go ahead and change it and then merge if you're okay with that.

PR #103, opened 2020-01-09
Proposed 10.0.0 Release

morethanfire:
This major release will rearrange the file structure using the model we have been talking about: Atoms, Molecules, Organisms, Templates, and Global. It will also fix all (or most of Eslint issues). 

The plan is to do this overtime and in a couple weeks we should have something ready for testing on HC. Hopefully with the file structure changes i don't believe there to be a huge regression risk in HC, but could be some caused by Eslint fixes. e.g. dangling underscores.


morethanfire:
I’m actually think of doing splitting this up into three different proposed versions: eslint fixes that won’t cause a regression, eslint fixes that could cause regression, and the file structure change. My belief with this new thought is to limit the regression risks into multiple version in hopes to be a more manageable job.

groberts314:
Yeah, that does sound like a good plan.  I haven't reviewed this yet, so if you'd like to make changes, go for it.

groberts314:
Plus right now there are 666 files changed; that cannot be a good sign! :-P

image

morethanfire:
Bwahaha. Yep we need to close this, we cannot have that amount of files changed.

Issue #25, opened 2018-06-19
Inconsistency across components with respect to "disabled" prop


groberts314:
*Do you want to request a feature or report a bug?*
Bug/Inconsistency

*What is the current behavior?*
The `Dropdown` and `PhoneInput` components have a `disable` prop while other components have `disabled` props.  All components should use `disabled` for consistency.

*If the current behavior is a bug, please provide the steps to reproduce.*
N/A

*Which version of react-cm-ui, and which browser / OS are affected by this issue? Did this work in previous versions of react-cm-ui?*
All released versions up to and including 1.11.11
Not browser or OS specific; it is an issue with the component prop definitions.

groberts314:
Cross reference in VSTS: [Backlog Item 10347](https://saddlebackchurch.visualstudio.com/DefaultCollection/Church%20Management/_workitems/edit/10347)

morethanfire:
I am going to open a PR with the work that adds the remaining components that need to deprecated `disabled` and add the `disable` prop.

PR #108, opened 2020-01-27
Form Fields and Grid Style/JS fixes

morethanfire:


src/elements/checkbox.js
```diff
@@ -28,10 +62,52 @@ class Checkbox extends Component {
...
+            }
+        } else {
+            event.stopPropagation();
+        }
+    }
+
+    onLabelClick(event) {
+        const { labelClick } = this.props;
+
+        if (!_.isUndefined(labelClick) && labelClick === false) {
```
groberts314: I think `labelClick === false` is sufficient here; checking it for `undefined` beforehand doesn't really add anything, does it?  NAB of course.

src/elements/checkbox.scss
```diff
@@ -79,12 +83,16 @@ $padding-label-toggle: 63px;
...
             cursor: auto;
-            &::before { background: color(backgroundColorDisable); }
-            .ui.icon-check { color: color(backgroundColorStatic); }
+            &::before {
+                background-color: color(backgroundColorDisable);
+                border: 1px solid color(borderColor);
+            }
+        }
+        &:checked + label .ui.icon-check {
+            opacity: 1;
```
groberts314: Do these changes by chance address that long-standing issue where disabled checkboxes would always appear to be checked?  If not, no biggie.  It looks though like it was perhaps getting into that area.  If so, I'm excited, and many thanks!

src/modules/phoneInput.js
```diff
@@ -154,7 +176,7 @@ class CountryDropdown extends React.PureComponent {
         return (
             <Dropdown
                 clearable={false}
-                disabled={disabled}
+                disable={disabled}
```
groberts314: Thanks for fixing that.  _Someday_ I'd like to fix `<Dropdown />` so it really is `disabled` (for consistency), but that's a larger effort.  See #25.

src/modules/phoneInput.js
```diff
@@ -227,6 +249,7 @@ class PhoneInput extends React.PureComponent {
                     className=""
                     countryOptions={countryOptions}
                     countrySelectComponent={CountryDropdown}
+                    fluid=""
```
groberts314: Why is `fluid` and empty string here?  Shouldn't it be a Boolean?

src/elements/checkbox.js
morethanfire: Yep, I agree.

src/modules/phoneInput.js
morethanfire: That is on purpose because ReactPhoneInput throws a warning. I think it's good for now, but really what should happen is removing the spreading of props.

src/modules/phoneInput.js
morethanfire: I agree.

morethanfire: We probably could change them all to `disable` and have a `disabed` alias for the time being.

src/elements/checkbox.scss
morethanfire: Yes

src/modules/phoneInput.js
groberts314: Do you preferred `disable` or `disabled`?  I prefer `disabled` myself.

src/modules/phoneInput.js
groberts314: Ok, fair enough.

src/modules/phoneInput.js
morethanfire: `disable`

morethanfire: Although, my original thought behind `disabled` is based on w3 specs. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#htmlattrdefdisabled

groberts314: Yes exactly ... I feel `disabled` is a bit more conventional.

groberts314: And I think for _most_ of React CM UI, we've used `disabled`.  I think `<Dropdown />` is an outlier (and that there's maybe one or two others also using `disable` but most use `disabled`).

PR #109, opened 2020-01-31
Page Fixes and New Background Color Prop

morethanfire:
This work is not be not breaking. I consider it a minor version to be able to sync up with design and be able to fix the problem in HC before it grows to large. 

However with that being said, it will cause a small amount of regressions due to how some of the Page sub components were being styled. I went through HC and found only 10 files that the Page component is being used and out of those ten there are only 5 locations that a regression would be caused. I packed these changes and installed on HC and fixes those 5 locations. That branch is now wait for this work to be released sot hat we can submit a PR in HC.


src/templates/content.scss
```diff
@@ -18,6 +24,15 @@ $prefixPageContent: page--content;
     }
 }
 
+@media only screen and (min-width: 508px) {
```
groberts314: NAB: Should we get the standardized breakpoints mixin setup in React CM UI as well?

morethanfire: I do think the mixins should come from this package, but for now I think we are okay, right?

groberts314: Right you are; hence NAB :-)

morethanfire: Yep, yep. Thanks!

PR #110, opened 2020-02-10
[dev] Tore out dropdown button out of dropdown

morethanfire:
This is 1 of 2 tasks to split apart the Dropdown component.


docs/src/atoms/dropdown.js
```diff
@@ -1,15 +1,25 @@
+/* eslint-disable */
+
+/**
+ * NOTE: The props 'button' part of this component's documentation has been is deprecated as of Feb. 11, 2020.
+ * Please use the DropdownButton component instead.
+ * We are disabling Eslint for this file because of the above note and wanting to
+ * deprecate it fully in hopes to rip out the select portion of it as well.
```
groberts314: Can I better understand why it is necessary to completely disable ESLINT?

docs/src/molecules/dropdownButton.jsx
```diff
@@ -0,0 +1,61 @@
...
+        {
+            name: 'as',
+            type: 'button',
+            default: 'button',
+            description: 'Can only be a button.',
+        }, {
+            name: 'iconType',
+            type: 'chevron-down<br />| plus',
+            default: 'chevron-down',
+            description: 'An element type to render as.',
```
groberts314: This description seems off for `iconType` prop (maybe it was accidentally copied and pasted from an `as` prop somewhere...?)

src/molecules/dropdownButton.jsx
```diff
@@ -0,0 +1,107 @@
...
+import React, { useRef, useState } from 'react';
+import { buttonPropTypes, buttonDefaultProps } from '../atoms/buttonConstants';
+import Button from '../atoms/button';
+import DropdownMenu from '../atoms/dropdownMenu';
+import DropdownMenuOption from '../atoms/dropdownMenuOption';
+import Icon from '../atoms/icon';
+
+const propTypes = {
+    ...buttonPropTypes,
+    as: PropTypes.oneOf(['button']),
```
groberts314: Why keep this is a `oneOf` (or even have it at all) if the only valid value is "button"?

docs/src/atoms/buttonIconExample.jsx
```diff
@@ -0,0 +1,23 @@
+import { Button, Icon } from 'react-cm-ui';
+import React from 'react';
+
+function ButtonIconExample() {
+    return (
+        <div>
+            <Button icon>
+                <Icon color="inverse" type="plus" />
```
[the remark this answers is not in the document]
morethanfire: Makes sense.

docs/package.json
```diff
@@ -5,6 +5,8 @@
     "build": "webpack",
     "build:clean": "rimraf build",
     "start": "webpack-dev-server --mode development --content-base docs/build/ --hot --progress --colors --host 0.0.0.0",
+    "start:local": "npm run start",
+    "start:local:prod": "npm run build",
```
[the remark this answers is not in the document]
morethanfire: I want to do more with the script that just build, but kind of what we do over in HC.

docs/src/atoms/dropdown.js
morethanfire: Because I don't see a need to fix the whole entire file when it's about to be fully deprecated.

docs/src/molecules/dropdownButton.jsx
morethanfire: Oh, you're right. I forgot to update it.

src/molecules/dropdownButton.jsx
morethanfire: You're right.

morethanfire: fixed.

docs/src/molecules/dropdownButton.jsx
morethanfire: fixed.

morethanfire:
I added a few more commits to this PR before I wanted to call it done. Apologies for that.

Added sense last review:
- A component
- Prompt's children renderer was reworked in order to work with any component. Thinking about DropdownButton.
- Color import fixes.
- Docs Dropdown button prop components were replaced with DropdownButton.
- A few misc. fixes around docs.

docs/src/atoms/promptDropdownButtonExample.jsx
```diff
@@ -0,0 +1,74 @@
...
+                <DropdownButton.Option
+                    id="edit"
+                    label="Edit"
+                    onClick={() => onPromptClick('edit')}
+                />
+                <DropdownButton.Option
+                    id="delete"
+                    label="Delete"
+                    onClick={() => onPromptClick('delete')}
+                />
```
groberts314: For the sake of the docs showing good coding practice, is it better if this becomes a class component, and has separate, bound functions for handling "delete" versus "edit", so as to avoid the anonymous functions in the render?  Just a thought.  NAB.

docs/src/atoms/promptExample.jsx
```diff
@@ -0,0 +1,26 @@
...
+        <div>
+            <Prompt inline>
+                <Button color="success">Save Me!</Button>
+            </Prompt>
+
+            <br />
+            <br />
+
+            <Prompt inline>
+                <A>asdf</A>
```
groberts314: Should this say something other than "asdf"?  "Click Me"?

docs/src/atoms/promptExample.jsx
```diff
@@ -0,0 +1,26 @@
+import {
+    A,
+    Button,
+    DropdownButton,
+    Prompt,
+} from 'react-cm-ui';
+import React from 'react';
+
+function PrompExample() {
```
groberts314: Misspelled `PromptExample` here (missing the "t" in "Prompt").

src/atoms/a.jsx
```diff
@@ -0,0 +1,51 @@
...
+        <span
+            className={containerClassName}
+            onClick={onClick}
+            onKeyDown={onKeyDown}
+            role="button"
+            tabIndex={tabIndex}
+            style={style}
+        >
+            {children}
+        </span>
```
groberts314: Very interesting... 🤔 
Are there any consequences (good or bad) to these actually being `<span>`s, rather than bona fide, 100% organic, good ol' fashioned `<a>` elements?

src/atoms/prompt.jsx
```diff
@@ -71,45 +107,61 @@ class Prompt extends Component {
...
+    //             ref: (promptAction) => { this.promptAction = promptAction; },
+    //         });
+    //     }
+
+    //     return React.cloneElement(children, {
+    //         className: `${promptActionClasses} ${show ? 'color-static' : null}`,
+    //         onClick: this.onClick,
+    //         ref: (promptAction) => { this.promptAction = promptAction; },
+    //     });
+    // }
```
groberts314: Can we please remove the commented code, once we commit to the current implementation?

src/atoms/prompt.jsx
```diff
@@ -140,60 +192,51 @@ class Prompt extends Component {
...
-                {this.renderAction()}
+            <div
+                className={containerClasses}
+                id={id}
+                style={style}
+            >
+                {/* {this.renderAction()} */}
+                <div
+                    onClick={this.onClick}
+                    onKeyDown={() => {}}
```
groberts314: Should we have a `const noop = () => {};` outside of `render()` and use that instead?  I thought anonymous functions inside the `render()` was a no-no (essentially as bad as `.bind(this)` inside `render()`).

src/atoms/prompt.jsx
```diff
@@ -140,60 +192,51 @@ class Prompt extends Component {
...
-                    {/* eslint-disable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions, jsx-a11y/anchor-is-valid */}
-                    <a className="prompt-yes-btn" onClick={this.onYesClick}>Yes</a>
-                    <a className="prompt-no-btn" onClick={this.onNoClick}>No</a>
-                    {/* eslint-enable jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions, jsx-a11y/anchor-is-valid */}
-                    {/* eslint-enable max-len */}
+
+                    <div
+                        className="prompt-yes-btn"
+                        onClick={this.onYesClick}
+                        onKeyDown={() => {}}
```
groberts314: Another case for my `noop` const

src/atoms/prompt.jsx
```diff
@@ -140,60 +192,51 @@ class Prompt extends Component {
...
+                        role="button"
+                        tabIndex={-1}
+                    >
+                        Yes
+                    </div>
+
+                    <div
+                        className="prompt-no-btn"
+                        onClick={this.onNoClick}
+                        onKeyDown={() => {}}
```
groberts314: Another case for my `noop` const

src/templates/actionBar.jsx
```diff
@@ -136,185 +164,217 @@ class ActionBar extends React.PureComponent {
...
+                                                    disable={dropdownButton.disable}
+                                                    iconType={dropdownButton.iconType}
+                                                    id={dropdownButton.id}
+                                                    label={dropdownButton.label}
+                                                    onClick={dropdownButton.onClick}
+                                                    style={dropdownButton.style}
+                                                    title={dropdownButton.title}
+                                                >
+                                                    {_.map(dropdownButton.options, (option) => {
+                                                        dropdownButtonOptionKeyNum += 1;
```
groberts314: Look at you not using unary `++` and then suppressing the warning! 😁 
I think I'm perhaps a bit more of a rebel...

src/templates/actionBar.jsx
```diff
@@ -136,185 +164,217 @@ class ActionBar extends React.PureComponent {
...
+                                                    })}
+                                                </DropdownButton>
+                                            )}
 
-                                                            {jsx}
-                                                        </List.Item>
-                                                    );
-                                                })}
-                                            </List>
+                                            {list && _.isArray(list) && (
```
groberts314: Is it necessary to do `list && _.isArray(list)`?  Shouldn't `_.isArray(list)` be sufficient?

src/templates/actionBar.jsx
```diff
@@ -136,185 +164,217 @@ class ActionBar extends React.PureComponent {
...
+                                                <List
+                                                    className="action_bar--list"
+                                                    horizontal
+                                                >
+                                                    {_.map(list, (item) => {
+                                                        const {
+                                                            actionsButton,
+                                                            jsx, // eslint-disable-line no-shadow
```
groberts314: Hmm, are we sure this is okay?  This is one where (despite my rebelliousness mentioned above) I'd probably heed the warning and maybe alias this as `innerJsx` or some such to avoid the variable shadowing.  That's one of those things in my book where the warning is correct in telling you you might be doing something dangerous; think twice.  If it's correct, and you want to leave it as-is, no objections.  NAB.

src/templates/actionBar.jsx
```diff
@@ -136,185 +164,217 @@ class ActionBar extends React.PureComponent {
...
+                                                            item.divide;
+
+                                                        const listItemClassName = ClassNames('action_bar--list_item', {
+                                                            'action_bar--list_item-jsx': jsx,
+                                                            'action_bar--list_item-icon_filter': iconFilter,
+                                                            'action_bar--list_item-icon_grid': iconGrid,
+                                                            'action_bar--list_item-icon_table': iconTable,
+                                                        });
+
+                                                        const itemKey = `action_bar--list_item-${gridColumnListItemKeyNum++}`; // eslint-disable-line no-plusplus
```
groberts314: Haha there it is! 😁 `++` FTW!

groberts314: `++` has been in so many programming languages using syntax for so long, I'm not sure when it officially because "considered harmful".  I think there are situations where you should probably avoid it (i.e. incrementing a variable and reading its value at the same time is just not smart there) but I think banning it wholesale is a bit harsh and unnecessary.

docs/package.json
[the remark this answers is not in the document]
morethanfire: Fixing.

docs/src/atoms/promptDropdownButtonExample.jsx
morethanfire: According to the React docs writing function components is the way they're want everyone to head. I think here, https://reactjs.org/docs/hooks-state.html

docs/src/atoms/promptExample.jsx
morethanfire: Thanks

docs/src/atoms/promptExample.jsx
morethanfire: Thanks!

src/atoms/a.jsx
morethanfire: It mostly has to do with people that have special needs, but as well the docs say not using an `href` with an a tag isn't semantically correct.

src/templates/actionBar.jsx
morethanfire: Oh, let me get this one.

src/atoms/prompt.jsx
```diff
@@ -140,60 +146,51 @@ class Prompt extends Component {
...
 
         return (
-            <div className={containerClasses} id={id} style={style}>
-                {this.renderAction()}
+            <div
+                className={containerClasses}
+                id={id}
+                style={style}
+            >
+                {/* {this.renderAction()} */}
```
groberts314: Is there a reason to keep this guy around?  NAB of course.

src/atoms/a.jsx
groberts314: Got it.  Yeah, thinking about this more, basically the warnings are telling us that we're kind of misusing `<a>` elements.  So, our response is, okay then, we won't use `<a>` elements for that; we'll use `<span>` elements with `role="button"` to declare that they're interactive elements.  Makes sense.

PR #111, opened 2020-02-18
Added basic DataGrid rowProps properties

morethanfire:


docs/src/templates/drawerDataGridSample.jsx
```diff
@@ -0,0 +1,166 @@
...
+        setIsDrawerOpen(!isDrawerOpen);
+    }
+
+    function onClickTest() {
+        console.log('You just clicked the fourth column!'); // eslint-disable-line no-console
+    }
+
+    return (
+        <div>
+            <Button onClick={() => onDrawerToggle()}>Open Drawer</Button>
```
groberts314: Silly question: Can this just be `<Button onClick={onDrawerToggle}>Open Drawer</Button>`?  Or does that not work?  If it does work, we should update other placements.  I'd say NAB since it's an issue in the Docs app, and not in actual components.

morethanfire:
Like this @raoufr https://github.com/saddlebackdev/react-cm-ui/pull/111/files#diff-9777ca8d2a868eaf19b91bd8bdde90d5R156 ?

morethanfire:
For an example, if I understand the question correctly, you could do this:

```
const foo = true;

<Drawer.DataGrid
    ...
    rowProps={(data) => ({
        className: foo ? 'block_name--element_name' : null,
    })}
/>

File: docs/src/templates/drawerDataGridSample.jsx [hunk 2]
```diff
@@ -0,0 +1,166 @@
...
+                />
+
+                <Drawer.ActionBar
+                    columns={[
+                        {
+                            list: [
+                                {
+                                    jsx: (
+                                        <Icon
+                                            onClick={() => {}}
```
groberts314: NAB.  We should probably define a `const noop = () => {};` and use that to avoid having to do this in `render()`.

src/templates/dataGridTable.jsx
```diff
@@ -95,7 +95,7 @@ class DataGridTable extends React.PureComponent {
         const containerClasses = ClassNames('ui', `${classNamePrefix}_table`, className);
         const bodyClasses = ClassNames({ [`${classNamePrefix}_drop_shadow`]: dropShadow });
         const isSelectable =
-            _.isFunction(rowProps) && _.isFunction(rowProps().onClick);
+            !_.isUndefined(rowProps) && _.isFunction(rowProps().onClick);
```
groberts314: Just curious: Any particular reason for this change.  Since it should either be a function or `undefined`, the prior version should have been okay too, right?

raoufr:
Thank you @morethanfire!

PR #115, opened 2020-03-07
New TimeFromNow Component

morethanfire:


src/utils/timeFromNow.jsx
```diff
@@ -0,0 +1,127 @@
+import ClassNames from 'classnames';
+import moment from 'moment-timezone';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const propTypes = {
+    className: PropTypes.string,
+    date: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.shape({}),
```
groberts314: I'm assuming it's really "one of type: string or moment object" ... right?  I'm wondering if we should consider a custom check function to verify that it is in fact a moment object ( using [`moment.isMoment()`](https://momentjs.com/docs/#/query/is-a-moment/) ), similar to [what I did for the `<DurationPicker />` component](https://github.com/saddlebackdev/react-cm-ui/blob/dev/src/molecules/durationPicker.js#L293-L301).  At the time I wrote the `<DurationPicker />` I did discover there is a [`react-moment-proptypes`](https://www.npmjs.com/package/react-moment-proptypes) package that has these built in that we can consider.

All of this is firmly in NAB territory of course.  Up to you if you want to code the custom prop type check function for a moment object or not.

src/utils/timeFromNow.jsx
```diff
@@ -0,0 +1,127 @@
...
+        moment.relativeTimeThreshold('s', relativeTimeThreshold.s || defaultThreshold_s);
+        moment.relativeTimeThreshold('m', relativeTimeThreshold.m || defaultThreshold_m);
+        moment.relativeTimeThreshold('h', relativeTimeThreshold.h || defaultThreshold_h);
+        moment.relativeTimeThreshold('d', relativeTimeThreshold.d || defaultThreshold_d);
+        moment.relativeTimeThreshold('M', relativeTimeThreshold.M || defaultThreshold_M);
+        /* eslint-enable camelcase */
+    }
+
+    // Setting fromNow string with customized relativeTime
+    const fromNowTime = moment(date).fromNow(true);
```
groberts314: Are we presuming that the `date` value has already been shifted to the user's present time zone?
I'm pretty sure `moment(date)`, if `date` is a string representing a date/time, and does not have timezone or UTC offset, will instantiate a moment in the current time zone of the machine it's executing on.  That's a recipe for possible time zone troubles.

This is not necessarily a problem; I'm just flagging it as a potential problem.  We'd want to do some testing to be sure.

morethanfire: Same here, as above, hopefully it'll just be solved by the propTypes though.

groberts314: Yeah, I think let's take a wait and see approach here.  If it ain't broke, don't fix it.  If we start having time zone blues, we'll dig in and see what's what.

PR #116, opened 2020-03-09
[dev] AB#25758 - RB: [Person Record 2.0] [page] [lg - xl] - Notes: Drawer’s max width needs to never be greater than the inside of the page gutter (All Browsers)

fantaJinMode:
PR contains

- Ability to change drawers width/height whenever `maxwidth` and `maxHeight` props are updated. Initially, maxWidth and maxHeight was being considered only on drawer open.


fantaJinMode:
> Looks pretty good to me.
> NAB: Do we have constants for the default max width and max height somewhere? If not, should we?

I believe, there isn't contants for default max width and max height. I can do those changes around constants. @morethanfire Could you please correct me if there any constants on `react-cm-ui`

PR #120, opened 2020-03-10
[dev] AB#25753 - [Person Record 2.0] [page] [md] - Notes: Need to use mobile notes drawer UX. (All Browsers) 

fantaJinMode:
PR contains:
- Ability to pass custom prop named `mobileMaxWidth`.  It will enable drawer to show stylings as mobile drawers for viewport less than `mobileMaxWidth`

eg: Person record notes drawer for viewport less than 1018px width, should show mobile styling for notes drawer(Default drawer for mobile styling is 767px)

Note: Right now, I am passing `mobileMaxWidth` props to `drawer`, `FiltersDrawer`, `FilterRails`. For future enhancement, I am thinking of passing only `mobileMaxWidth` props to `drawer`. It can pass it to its children (FiltersDrawer, FilterRails). For that, I was thinking of using `React.context`. Please provide any suggestions/ideas for it.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "reacm-cm-ui-docs",
-  "version": "0.2.0",
+  "version": "0.3.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "reacm-cm-ui-docs",
+  "name": "react-cm-ui-docs",
   "version": "0.3.1",
   "lockfileVersion": 1,
   "requires": true,
--- a/src/templates/drawer.jsx
+++ b/src/templates/drawer.jsx
@@ -29,7 +29,6 @@ const propTypes = {
         PropTypes.number,
         PropTypes.string,
     ]),
-    mobileMaxWidth: PropTypes.number,
     onClickOutside: PropTypes.bool,
     onClose: PropTypes.func,
     onCloseComplete: PropTypes.func,
@@ -48,7 +47,6 @@ const defaultProps = {
     dimmer: true,
     maxHeight: undefined,
     maxWidth: undefined,
-    mobileMaxWidth: undefined,
     onClickOutside: undefined,
     onClose: undefined,
     onCloseComplete: undefined,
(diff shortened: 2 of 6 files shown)


docs/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reacm-cm-ui-docs",
```
groberts314: We should fix this typo (in parent `package.json` file of course): `react-cm-ui-docs`
First time I've noticed that! :-)
Definitely a NAB for this PR ... it's been broken a while now.

src/templates/drawer.jsx
```diff
@@ -254,6 +259,7 @@ class Drawer extends React.Component {
         const layeredOffset = 11;
         const scrollPosition = window.pageYOffset;
         const zIndex = 10002; // adding 2 accounts for the frist .drawer and .drawer-dimmers- z-indexes
+        const drawerWidth = mobileMaxWidth ? mobileMaxWidth + 1 : DEFAULT_DRAWER_WIDTH; // This implies if mobileMaxWidth is set and maxWidth not set
```
groberts314: Curious ... why + 1?

src/templates/drawer.jsx
```diff
@@ -416,7 +423,7 @@ class Drawer extends React.Component {
                             autoHide
                         >
                             <div className="drawer-container-inner">
-                                {children}
+                                    {children}
```
groberts314: Looks like we added extra whitespace here?  Any reason?  Could we revert that?

src/templates/filtersDrawer.jsx
```diff
@@ -238,12 +241,13 @@ class FiltersDrawer extends React.Component {
 
         return (
             <MediaQuery
-                maxWidth={767}
+                maxWidth={mobileMaxWidth ? mobileMaxWidth : 767}
```
groberts314: Could we do `mobileMaxWidth || 767`?

src/templates/drawer.jsx
fantaJinMode: Like if, `mobileMaxWidth=1018` then drawer width would be around 1018. Previously, it was handled like this for 767.

src/templates/drawer.jsx
fantaJinMode: Thanks @groberts314 , for catching it. Fixed.

src/templates/filtersDrawer.jsx
fantaJinMode: Fixed

docs/package-lock.json
fantaJinMode: Fixed

groberts314: I fixed it in `/docs/package.json` in b3a5970 =)

morethanfire:
:shipit:

PR #123, opened 2020-03-12
AB#25772 Fixes DrawerDeprecated Regression

morethanfire:


src/templates/drawerDeprecated.js
```diff
@@ -461,6 +461,7 @@ class DrawerDeprecated extends Component {
         const animationEvent = this._transitionProps(this.drawerContainerRef);
         const body = document.body;
         const drawerLength = document.querySelectorAll('.ui.drawer').length;
+        const drawerDeprecatedLength = document.querySelectorAll('.ui.drawer.deprecated').length;
```
groberts314: NAB: I'd rename these variables to indicate that these are a _count_ or _number_ of drawers and deprecated drawers (even if they are powered by the `.length` property of an array of DOM elements matching the selector).

PR #129, opened 2020-04-08
Upping the Performance of the Drawer's Animation.

morethanfire:
No rush in reviewing this and cutting a release. This work was part of a, "I hope this will fix the bug, but most likely won't." It doesn't fix the bug, but we might as well not lose the work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "moment-timezone": "^0.5.21",
     "prop-types": "^15.7.2",
     "react": "^16.8.6",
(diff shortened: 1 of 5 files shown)


morethanfire:
Yeah, no need for any polyfills, because all modern browsers support `translate3d`. We live in good time considering where browsers were, honestly, not too long ago.

groberts314:
True that!  Thanks!  :shipit: at will.

PR #130, opened 2020-04-08
Input required prop fixes.

morethanfire:
This is fore the church_management's PR [3329](https://github.com/saddlebackdev/church-management/pull/3329). 

Changes: 
- Added `componentDidUpdate` to set a new `showRequiredIndicator` state and to rerender the component with that value.
- Fixed majority of the Eslint errors.

Nothing in this should cause a regression except for maybe the `componentDidUpdate` function. Just to be aware of.


src/atoms/input.jsx
```diff
@@ -324,88 +252,232 @@ class Input extends React.PureComponent {
...
+                newType = 'tel';
+                break;
+            case 'text':
+            default:
+                newType = 'text';
+                break;
+        }
+
+        return newType;
+    }
```
groberts314, line 295: Seems to me that, in theory, we shouldn't really need this, unless consumers are passing invalid `type` props.

morethanfire: Should remove? Is it a blocker?

groberts314:
Testing Microsoft Teams integration.  This is a comment on a PR in `react-cm-ui` repo.

src/atoms/input.jsx
groberts314: Nah ... definitely a NAB.  Up to you whether to remove now (and possibly incur small, random regressions if consumers are using non-standard values).  Maybe safer to keep it.  I imagine at some point there was a good reason for it.  Those React CM UI consumers can be sloppy coders sometimes! 😁

morethanfire:
Oh, no! MS Teams!

PR #133, opened 2020-04-22
Validating drawer header exists before trying to modify its padding size

War777:
This fix must ensure that **_onHeaderUpdate** function validates the drawer header exists before modifiying its padding.

[azure devops ticket here](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/28059/)


src/templates/drawerDeprecated.js
```diff
@@ -147,13 +147,12 @@ class DrawerDeprecatedHeader extends Component {
 
     _onHeaderUpdate(node) {
         const headerEl = ReactDOM.findDOMNode(this);
-
         if (headerEl) {
             const paddingTop = headerEl.offsetHeight;
-
-            if (paddingTop !== this._drawerContainerInnerPaddingTop) {
+            let drawerInnerContainer = headerEl.closest('.drawer-container-inner');
```
sircris, line 152: I suppose that `drawer-container-inner` is a standard class for the container? Is this a safe assumption?

sircris:
Deferring approval and final review to @morethanfire

src/templates/drawerDeprecated.js
War777: Yes, it is an standard, the container, element with class it's created inside the render method, I suggest to move the class name to a constant in a separate task.

morethanfire: Yes, that is a safe assumption. It looks that `drawerInnerContainer` is not being redefined and should be a `const`.

War777: Done! Commit pushed :) thanks.

PR #141, opened 2020-05-19
Documentation Navigation Updates

morethanfire:


docs/src/app/navigation.jsx
```diff
@@ -10,16 +10,29 @@ import {
     colorInverse,
     colorStatic,
 } from 'react-cm-ui/styles/colorExports';
+import _ from 'lodash';
```
groberts314, line 13: I wonder if we should try and ["fix" all the Lodash imports](https://github.com/saddlebackdev/church-management/pull/3456#discussion_r424146737) in React CM UI at least, and then gradually adopt the practice as we fix bugs and do new development in HC (church-management).

docs/src/app/navigation.jsx
```diff
@@ -121,6 +134,7 @@ class CoreAppNavigation extends React.PureComponent {
 
         this.onLogoClick = this.onLogoClick.bind(this);
         this.toggleNavigation = this.toggleNavigation.bind(this);
+        this.listJSX = this.renderList();
```
groberts314, line 137: This is maybe a good idea for all of these arrays of columns and such, as long as they are indeed static/constant (i.e. they do not change based on props or state, etc.)  Saves having to "recompute" them on every render.

docs/src/app/navigation.jsx
```diff
@@ -157,12 +171,139 @@ class CoreAppNavigation extends React.PureComponent {
...
+
+                            return this.renderListItem(
+                                sectionItem,
+                                path,
+                                key,
+                            );
+                        })}
+
+                        {_.map(section.subSections, (subSection) => {
+                            sectionItemKeyNum += 1;
```
groberts314, line 221: Doesn't at least part of you wanna rebel and do `sectionItemKeyNum++; // eslint-disable-line no-plusplus` ...? ;-)

PR #148, opened 2020-05-28
[dev] AB#23667 - Banner and Z-Index Changes

morethanfire:


src/dataDisplay/banner.jsx
```diff
@@ -1,43 +1,198 @@
...
+    onAfterClose: undefined,
+    onOpen: undefined,
+    style: undefined,
+    title: undefined,
+    topPosition: undefined,
+    type: undefined,
+};
 
-class Banner extends Component {
+class Banner extends React.Component {
```
groberts314, line 66: Should it be a `PureComponent`?  Or too much regression risk in making that change?

src/dataDisplay/banner.jsx
```diff
@@ -1,43 +1,198 @@
...
+        } = this.state;
+
+        if (isOpen) {
+            this.onOpen();
+        }
+    }
+
+    componentDidUpdate(prevProps) {
+        const {
+            isOpen: prevIsOpn,
```
groberts314, line 92: `prevIsOpn` should be `prevIsOpen`.
I kind of like the convention `wasOpen` for the previous one, and keeping `isOpen` (i.e. don't alias it when you destructure it) for the current one, but the `next` and `prev` prefixes works too (and generally, not just for the special case of `isOpen`).

src/dataDisplay/banner.jsx
IlyaRadinsky: @groberts314  If we want to use the `PureComponent` - no any on-fly objects/functions should be used in `render()` method. But we have such style object, so, not sure that we have to rewrite the `render()` at the moment.

src/styles/theme.js
```diff
@@ -9,6 +9,12 @@ const theme = {
         keys: breakpointsKeys,
         values: breakpointsValues,
     }),
+    zIndex: {
+        drawer: 10000,
+        modal: 11000,
+        prompt: 12000,
+        banner: 13000,
+    },
```
IlyaRadinsky, line 17: These constants look different than the `variables.scss` constants.

src/dataDisplay/banner.jsx
groberts314: Roger that.  Thanks for setting me straight! :-)

src/styles/theme.js
morethanfire: Correct

morethanfire: The only difference is prompt and banner. Really, I don't think we should have prompt in there, but for now it is and should be higher than drawer and modal, but lower than banner.

groberts314:
@more answered @IlyaRadinsky 's question, and in order to unblock things and hopefully fix broken things, I think let's go ahead and merge this, despite the 'changes requested' status.

PR #150, opened 2020-06-03
Added and Fixed Data Type - Person Icons

morethanfire:


src/dataDisplay/icon/icon.jsx
```diff
@@ -2994,27 +3131,21 @@ class Icon extends React.PureComponent {
...
+                        pathId={pathId}
+                        renderGradientColor={renderGradientColor}
+                        style={svgStyle}
+                        title={title}
+                        type={type}
+                        viewBox="0 0 24 24"
+                    />
```
groberts314, line 3148: I wonder if there is some better pattern that we could apply here.  Like a factory or builder type pattern.  Just musing aloud ... NAB for sure.  Let's move along and unblock things!

PR #156, opened 2020-06-12
[dev] AB#30853 - [My Events] - Custom fields drawer behind edit event drawer

morethanfire:


.eslintrc
```diff
@@ -58,5 +58,21 @@
...
+            }
+        }, {
+            "files": [
+                "*constants.js",
+                "*Constants.js"
+            ],
+            "rules": {
+               "import/prefer-default-export": "off"
+            }
+        }
```
groberts314, line 76: I'm digging sensible ESLINT overrides for sure! 👍

src/templates/drawer.jsx
```diff
@@ -277,7 +280,7 @@ class Drawer extends React.Component {
         } = this.props;
         const animationEvent = domUtils.cssTransitionType(this.drawerContainerRef);
         const boxShadowPositionX = this.isPositionX('right') ? '-' : '';
-        const drawerLength = document.querySelectorAll('.ui.drawer').length;
+        const drawerLength = document.querySelectorAll('.ui.drawer-is_modal').length;
```
groberts314, line 283: Should we consider renaming `drawerLength` to something like `numberOfModalDrawers`?
Yeah, it was calculated from the `.length` property of an array, but what's important to understanding the logic is that it represents a _count_ of--or _quantity_ of--things that we're interested in, right?  So in my opinion, this rename would improve readability of the code slightly.

src/templates/drawerDeprecated.js
```diff
@@ -459,8 +478,8 @@ class DrawerDeprecated extends Component {
         const { onCloseComplete, position, onClickOutside } = this.props;
         const animationEvent = this._transitionProps(this.drawerContainerRef);
         const body = document.body;
-        const drawerLength = document.querySelectorAll('.ui.drawer').length;
-        const drawerDeprecatedLength = document.querySelectorAll('.ui.drawer.deprecated').length;
+        const drawerLength = document.querySelectorAll('.ui.drawer-is_modal').length;
```
groberts314, line 481: If we elect to rename the corresponding variable in the other file, let's please do it here as well.

src/templates/drawerDeprecated.js
```diff
@@ -459,8 +478,8 @@ class DrawerDeprecated extends Component {
...
+        const drawerDeprecatedLength = document.querySelectorAll('.ui.drawer.deprecated.drawer-is_modal').length;
```
groberts314, line 482: And a similar rename here too please.  Feel free to push back and tell me you don't want to rename any of them! :-)

src/templates/drawer.jsx
groberts314: Rename looks good; resolving.

src/templates/drawerDeprecated.js
groberts314: Renamed; resolving

src/templates/drawerDeprecated.js
groberts314: Renamed; resolving.

PR #157, opened 2020-06-12
[dev] BUG AB#27358 [Events Central] - Please remove Phone from the Name column [drill through My Ministry tab]

IlyaRadinsky:
The issue is that when `DataGrid` has no data the header cell widths are not aligned and we can see the columns of the "sticked table" and the columns of the "scrollable table" at same time. So, this PR should fix that.


morethanfire:
:shipit:

PR #160, opened 2020-06-19
[dev] AB#31150 - Eslint Fixes

morethanfire:
Used `eslint ./src --fix`


docs/src/dataDisplay/banner.js
```diff
@@ -421,31 +420,31 @@ export default class ViewsBanner extends React.Component {
...
                 message: 'A short notification description',
                 title: `AC Banner ${id + 1}`,
-                type: 'notification'
+                type: 'notification',
             };
             acData.push(newAcData);
         }
 
-        this.setState({ acData: acData });
+        this.setState({ acData });
```
groberts314, line 447: Probably doesn't matter terribly (it's just to power a documentation sample), but I don't think we're following a proper immutability approach around `acData` value in state.

Maybe something like this...?

```JavaScript
_onAcBannerClick(id) {
    this.setState((prevState) => {
        const { acData: prevAcData } = prevState;
        const prevAcDataItem = _.find(prevAcData, (d) => d.id === id);
        let newAcData = null;
        
        if (_.isObject(prevAcDataItem)) {
            // if `prevAcDataItem` is a single object we want a new object with `isOpen` toggled
            newAcData = {
                ...prevAcDataItem,
                isOpen: !prevAcDataItem.isOpen,
            };
        } else {
            // if `prevAcDataItem` is an array, append a new item
            const newAcDataItem {
                id: id,
                id,
                isOpen: true,
                message: 'A short notification description',
                title: `AC Banner ${id + 1}`,
                type: 'notification'
                type: 'notification',
            };
            
            newAcData = [
                ...prevAcDataItem, // existing items in the array
                newAcDataItem,
            ];
        }
        
        return { acData: newAcData };
    });
}

```

We can consider this a NAB for the purposes of an ESLINT auto-fix PR and flag it and come back to it later (if at all).

docs/src/dataDisplay/banner.js
```diff
@@ -469,20 +468,17 @@ export default class ViewsBanner extends React.Component {
...
-                    key={`banner-key-${d.id}`}
-                    message={d.message}
-                    onAfterClose={this._onAfterClose.bind(this)}
-                    onClose={this._onAcBannerClick.bind(this, d.id)}
-                    title={d.title}
-                    type={d.type}
-                />
-            );
-        })
+        return _.map(this.state.acData, (d) => (
```
groberts314, line 471: TOTAL NAB:
I wonder if we should consider fixing up Lodash imports in React CM UI as well.
Maybe if/when we upgrade Webpack, so that the optimizer can do its job properly.

docs/src/global/breakpointActions.js
```diff
@@ -1,11 +1,10 @@
...
 
-import { appReduxStore } from '../global/configureReduxStore.js';
+import { appReduxStore } from './configureReduxStore.js';
 
 export default class BreakpointActions {
     static update() {
-        const clientHeight = document.documentElement.clientHeight;
-        const clientWidth = document.documentElement.clientWidth;
+        const { clientHeight } = document.documentElement;
+        const { clientWidth } = document.documentElement;
```
groberts314, line 7: Wouldn't we want `const { clientHeight, clientWidth } = document.documentElement;`?
Probably ESLINT auto fix is not quite smart enough to have been able to do that. :-)

docs/src/global/breakpointReducer.js
```diff
@@ -27,7 +27,7 @@ const breakpoint = (state = DEFAULT_STATE, action) => {
                 isXSmall: action.result.isXSmall,
                 isXXSmall: action.result.isXXSmall,
                 winWidth: action.result.winWidth,
-            });
+            };
```
groberts314, line 30: Nice!  That's a handy thing ESLINT can auto-fix.  I've been doing it manually.  Getting good at it, but I'm likely no match for ESLINT auto fix! 😁

docs/src/organisms/accordion.js
```diff
@@ -831,7 +839,7 @@ export default class CollectionsAccordion extends React.Component {
                     <Accordion selected={controlledExampleSelectedIndex}>
                         <Accordion.Item>
                             <Accordion.Summary onClick={this._onClickAccordionItem}>
-                                <Header size="small">{'Option One'}</Header>
+                                <Header size="small">Option One</Header>
```
groberts314, line 842: This is what I was talking about [here](https://github.com/saddlebackdev/react-cm-ui/pull/147#discussion_r442274597).

docs/src/organisms/card.js
```diff
@@ -313,41 +314,48 @@ export default class ViewsTitleBar extends React.Component {
...
+                            <p>
+                                Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec ultrices non diam vitae blandit.
                                 Curabitur quis velit eu nibh cursus convallis ac in arcu. Aenean malesuada sed leo eget cursus.
                                 Nunc urna magna, sodales eget rutrum et, facilisis id velit. Pellentesque eu ornare tellus.
                                 Pellentesque aliquet a nibh consectetur vestibulum. Aenean in ipsum tincidunt, congue ante a,
                                 dictum nunc. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos
                                 himenaeos. Aenean in tortor porttitor, tristique eros convallis, viverra nunc. Morbi dignissim
-                                sapien ac dui convallis molestie. Donec consequat odio in egestas lacinia.</p>
+                                sapien ac dui convallis molestie. Donec consequat odio in egestas lacinia.
+                            </p>
```
groberts314, line 358: Nice!  ❤️ ESLINT
I've been known to do this kind of thing manually as well ;-)

docs/src/organisms/titleBar.js
```diff
@@ -122,77 +122,71 @@ export default class ViewsTitleBar extends React.Component {
...
                     <Highlighter customStyle={{ marginBottom: '44px', marginTop: '44px' }}>
                         {subTitleSample}
                     </Highlighter>
 
                     {/* Children */}
-                    <Header size="large" style={{ marginTop: '55px' }} sub={true}>
+                    <Header size="large" style={{ marginTop: '55px' }} sub>
                         Children
                         <Header.Subheader>
                             Title Bar's can return it's children on the right side of the bar.
```
groberts314, line 168: Grammar nitpick: "its" (no apostrophe) is the possessive form of "it".  "it's" with an apostrophe is a contraction for "it is".
https://www.grammarly.com/blog/its-vs-its/

docs/src/templates/drawerActionBar.js
```diff
@@ -323,7 +325,10 @@ class ModulesDrawerActionBar extends React.Component {
                         </Header.Subheader>
                     </Header>
 
-                    Optional and Required Properties for <code>columns</code>:
+                    Optional and Required Properties for
+                    {' '}
+                    <code>columns</code>
+                    :
```
groberts314, line 331: This one (ESLINT insisting on all these things on separate lines) drives me a little bit nuts sometimes.  But c'est la vie.

src/templates/modal.js
```diff
@@ -422,13 +423,13 @@ class Modal extends Component {
     _onScrollStart() {
         const scrollContainerPos = this.modalContainerInner.parentNode.scrollTop;
 
-        this.setState({ isScrolled: scrollContainerPos > 0 ? true : false });
+        this.setState({ isScrolled: scrollContainerPos > 0 });
```
groberts314, line 426: I like that it can take care of this kind of thing. 👍

docs/src/organisms/titleBar.js
groberts314: Probably don't need "'s" after "Title Bar" either (Should be "Title Bar can return its children..." versus "Title Bar's can return it's children...") ... all sorts of bad grammar there! 😁

docs/src/atoms/header.js
```diff
@@ -1,8 +1,9 @@
-'use strict';
 
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Card, Header, Icon, TitleBar } from 'react-cm-ui';
+import {
+    Card, Header, Icon, TitleBar,
+} from 'react-cm-ui';
```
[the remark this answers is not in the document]
morethanfire: Yeah, I do wish that eslint would stack them, because that what the rule is actually wanting.

docs/src/atoms/label.js
```diff
@@ -300,5 +299,4 @@ export default class ElementsLabel extends React.Component {
     _onClearClick() {
         console.log('Label removed.');
```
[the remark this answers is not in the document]
morethanfire: Right.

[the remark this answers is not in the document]
morethanfire: `--fix` only fixes code that shouldn't regress anything.

docs/src/dataDisplay/banner.js
morethanfire: Yeah, I would agree that this is NAB for this particular problem.
Auto fix does help with removing quite a few issues, which i think is sort of convenient, beside not returning destructured props to their own lines.

docs/src/dataDisplay/banner.js
morethanfire: Yep, I think that would be a good idea that would be easy to implement.

docs/src/global/breakpointActions.js
morethanfire: Must not be smart enough. :/

groberts314:
@morethanfire Please check build failures.  Please run `npm run build` for the UI docs locally and see if you see the same errors as those reported by the build server (see https://dev.azure.com/saddlebackchurch/Church%20Management/_build/results?buildId=17949&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=7cc137c2-61f2-5d65-b5f8-0e03d1af3170 ).

docs/src/dataDisplay/banner.js
groberts314: I kind of like the approach of doing it while we're doing other work on various files.  It's not too hard.  Obviously mistakes can be made that don't get caught during code review and escape into released code (this is why we also need to be doing unit tests! :-) so there is risk of regression.  But I think it's worth it to take some risk if, in the long term, this will help us optimize.

morethanfire:
The build failure have been resolved in commit 1cf52b1.

groberts314:
I see this is now in `9.14.0` milestone (not vNext === `0.9.13`) so I am going to add **`do not merge`** label as a precaution.

docs/src/dataDisplay/banner.js
groberts314: Resolving.

groberts314:
Removing **`do not merge`** label since we have released `9.13.0` and vNext is now `9.14.0`.

PR #162, opened 2020-06-22
[dev] AB#24428 & AB#24723 - React-CM-UI Documentation Components

morethanfire:


morethanfire:
Oh! I like the milestone you added @groberts314

docs/src/global/componentApi.jsx
```diff
@@ -0,0 +1,96 @@
...
+        margin: '33px -22px 0',
+    },
+});
+
+function ComponentApi(props) {
+    const {
+        docs,
+    } = props;
+    const classes = useStyles();
+    const isDocsNotDefined = isNil(docs) || isEmpty(docs);
```
groberts314, line 42: I think `isEmpty()` covers `isNil()` (i.e. it will return `true` for `undefined` or `null` as well as empty strings and arrays and objects) ... so we can probably just use `isEmpty()`.

morethanfire: 👍

morethanfire:
@groberts314 do we have a shipit?

docs/src/global/componentApi.jsx
groberts314: Confirmed this is fixed in 1c3b162.  Resolving.

groberts314:
:shipit:

PR #169, opened 2020-06-24
[dev] AB#27368 - Page gutters should be 44px at the desktop breakpoint (768px+)

morethanfire:
- Small (375px)
  - Page Gutters: 22px
- Special Case Break Point (496px) and Up
  - Page Gutters: 44px


docs/src/templates/pageDemo.jsx
```diff
@@ -1159,4 +1179,11 @@ const mapStateToProps = (state) => {
     };
 };
```
groberts314, line 1180: Should we move `mapStateToProps()` up to the top, below the `propTypes` and `defaultProps` declarations while we're working in here?

groberts314:
PR Review Checkpoint Meeting Thurs 06/25/2020.
:shipit:

PR #170, opened 2020-06-25
[dev] Cleaned up Icon And Added Ability To Suppress Title

### LINKED ISSUE #34: Add ability to suppress Icon title/tooltip
*Do you want to request a feature or report a bug?*
Technically this is a new feature request, but it's also kind of a bug fix too, as the current behavior is kind of undesirable and likely wasn't intentional.

*What is the current behavior?*
`<Icon>`s currently _always_ show a title/tooltip on hover.  The displayed value defaults to the Icon's type name, which is what currently shows in 99.9% of Icon placements in the Healthy Church app! (Disclaimer: percentage made up on the spot and very likely exaggerated for dramatic effect, but you get the idea :-)  fdfec09 added the ability to set the desired title/tooltip via a new prop, which is nice, but perhaps not yet the final solution.  I'm proposing that we add an additional prop to optionally suppress the title/tooltip altogether (assuming it is valid for SVGs to not have a `<title>` element, or perhaps using some attribute or other mechanism to request that browsers not use the title element to show a tooltip on hover), since there are many places in the UI where the tooltip feels unnecessary and even inappropriate.

*If the current behavior is a bug, please provide the steps to reproduce.*
Hover over an icon to reveal the tooltip.

*Which version of react-cm-ui, and which browser / OS are affected by this issue? Did this work in previous versions of react-cm-ui?*
* React CM UI version 1.16.3 (the latest version as of filing this bug) and most previous versions are affected.
* The ability to set the title prop (commit fdfec09 referenced above) was introduced in 1.13.4.
* I am using Google Chrome Version 70.0.3538.77 (Official Build) (64-bit) on Windows 7 but I don't believe the issue is particularly browser or OS dependent (unless different browsers will show or not show a tooltip on hover using the SVG's `<title>` element; I haven't tried any other browsers).

morethanfire:
Fixes #34.


src/dataDisplay/icon/iconSVG.jsx
```diff
@@ -1,55 +1,107 @@
...
+    }
+
+    if (polygon) {
+        id = polygon.id;
+    }
+
+    if (type === false) {
+        console.log('## type', type);
+        console.log('id', id);
+    }
```
groberts314, line 72: Do we need these log messages?  If so, suppress ESLINT please.  Otherwise please remove them.

src/dataDisplay/icon/iconSVG.jsx
```diff
@@ -1,55 +1,107 @@
...
+
     return (
         <svg
             style={style}
             viewBox={viewBox}
         >
-            <title>{title || type}</title>
+            {title !== false && <title>{title || type}</title>}
```
groberts314, line 79: Very nice! 👍

groberts314:
PR Review Checkpoint Meeting Thurs 06/25/2020.
Cameron to make small fix and then merge it.

src/dataDisplay/icon/iconSVG.jsx
morethanfire: Cleaned up. Thanks!

groberts314: Yep, looks like they're gone.  Resolving.

morethanfire:
Thanks @groberts314!

PR #187, opened 2020-08-05
[dev] AB#32319 AB#32320 AB#32321 AB#32322 - Reorganized Components In `./docs`

morethanfire:


docs/src/dataDisplay/headerDeprecated/headerDeprecated.jsx
```diff
@@ -6,10 +6,10 @@ import {
 } from 'react-cm-ui';
```
groberts314, line 6: The component is still `<Header>` not `<HeaderDeprecated>` ... are there near future plans to deprecate it?

docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
```diff
@@ -17,10 +17,10 @@ import { Card, Dropdown, Grid, Header, SubNavigation, TitleBar } from 'react-cm-
 import { Link } from 'react-router';
```
groberts314, line 17: Are we planning on renaming `<Dropdown>` => `<DropdownDeprecated>`?

docs/src/layout/containerDeprecated/containerDeprecated.jsx
```diff
@@ -0,0 +1,48 @@
+import {
```
groberts314, line 1: Are we planning on changing `<Container>` to `<ContainerDeprecated>`?
This deprecation I think I at least expected ( `<Header>` and `<Dropdown>` caught me a little bit off-guard).

docs/src/navigation/subNavigationDeprecated/subNavigationDeprecated.js
```diff
@@ -1,15 +1,13 @@
-
 import MediaQuery from 'react-responsive';
```
groberts314, line 1: Are we planning on renaming the component to indicate it is deprecated?

docs/src/routes.jsx
```diff
@@ -11,495 +16,39 @@ const routes = (
...
+                )) : map(levelOne.items, (levelOneItem) => (
+                    <Route
+                        getComponent={(location, callback) => {
+                            import(`./${levelOne.directory}/${levelOneItem.component}`)
+                                .then((module) => callback(null, module.default));
+                        }}
+                        key={`${levelOneItem.path}`}
+                        path={levelOneItem.path}
+                    />
+                ))}
```
groberts314, line 49: Nice and clean!  I like!

docs/src/dataDisplay/headerDeprecated/headerDeprecated.jsx
morethanfire: Yeah, it is a soft deprecation; bwahaha!
We should only be using the Typography component now.

docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
morethanfire: I don't think we should rename these in `./src`, but in `./docs/src` my thinking was/is to better identify what is deprecated. I could be wrong. Let me know.

docs/src/layout/containerDeprecated/containerDeprecated.jsx
morethanfire: I don't think we should rename these in `./src`, but in `./docs/src` my thinking was/is to better identify what is deprecated. I could be wrong. Let me know.

docs/src/navigation/subNavigationDeprecated/subNavigationDeprecated.js
morethanfire: Hard deprecation. We should no longer be using this from here and out.

I don't think we should rename these in `./src`, but in `./docs/src` my thinking was/is to better identify what is deprecated. I could be wrong. Let me know.

docs/src/routes.jsx
morethanfire: Yeah, I believe this will help with maintenance.

PR #190, opened 2020-08-17
[dev] PBI AB#32635 Ability to create an event - When & Where - [E.1.2] - Recurring Event Re-Design

IlyaRadinsky:
These changes will allow to have a different displaying format of not focused `DatePickerInput` (displayFormat = 'dddd MM/DD/YYYY'):

Focused:
Screen Shot 2020-08-17 at 12 22 36 PM

Not focused:
Screen Shot 2020-08-17 at 12 22 24 PM

But here is one problem - at the moment `Input` control has inconsistent display flow. On initialization step it shows the `props.value` but then if you input any string it writes a new value directly into `input` control and shows that value. In this case the `Input` will not reflect on any changes of the `props.value`.  
@morethanfire  I see that initially you tried to use internal state to store the changed string, why did you give up this idea? Does it make sense to go back to this option?

### DIFF
--- a/src/inputs/datePickerInput/datePickerInput.jsx
+++ b/src/inputs/datePickerInput/datePickerInput.jsx
@@ -181,7 +181,7 @@ class DatePickerInput extends React.PureComponent {
         }
     }
 
-    onCalendarChange({ date, dateFrom, dateTo }, isFocused = false) {
+    onCalendarChange({ date, dateFrom, dateTo }) {
         const {
             disable,
             disabled,
@@ -203,11 +203,11 @@ class DatePickerInput extends React.PureComponent {
             });
         }
 
-        this.setOpen(false, isFocused);
+        this.setOpen(false);
     }
 
     onCalendarClickOutside() {
-        this.setOpen(false, false);
+        this.setOpen(false);
     }
 
     onIconClick() {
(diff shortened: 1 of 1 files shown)


groberts314:
PR Review Checkpoint Meeting Tues 2020-08-18.
We need to understand whether this is critical to get into HC 1.27.2 release or not.  Depending on that, we may have to cherry-pick to patch 9.13.x React CM UI as well as current version.

src/inputs/datePickerInput/datePickerInput.jsx
```diff
@@ -264,8 +271,13 @@ class DatePickerInput extends React.PureComponent {
         }
     }
 
+    onInputClick() {
+        this.setState({ isFocused: true });
+    }
+
     onInputFocus() {
         this.setOpen(true);
+        this.setState({ isFocused: true });
```
morethanfire, line 280: Why are we setting `isFocused` to `true` and above in the `onInputClick()` method?

IlyaRadinsky: So, we click on control (or control got focus) and able to edit the text - in this case we set the `isFocused` flag and do not use the `displayFormat`.

groberts314:
PR Review Checkpoint Meeting Thurs 2020-08-20.
Cameron to follow up on this.  There are still questions and concerns,

morethanfire:
@IlyaRadinsky 
> But here is one problem - at the moment Input control has inconsistent display flow. On initialization step it shows the props.value but then if you input any string it writes a new value directly into input control and shows that value. In this case the Input will not reflect on any changes of the props.value.
@morethanfire I see that initially you tried to use internal state to store the changed string, why did you give up this idea? Does it make sense to go back to this option?

We went away from the value being stored in state because we were getting warnings about going from uncontrolled to controlled. Now, the component is definitly not a uncontrolled component, but the current fix does help with removing those warnings. What is your suggestion with know this?

Good to know, if this input fix is not urgent, we could classify this as a UI input gap work on this issue for 10.0.0.

PR #193, opened 2020-08-25
[release/9.15.1] Bug AB#33063 Fix international option of country options

raoufr:


src/inputs/phoneInput/phoneInputCountryDropdownItem.jsx
```diff
@@ -70,9 +70,9 @@ class PhoneInputCountryDropdownItem extends React.Component {
         return (
             <div
                 className={containerClasses}
-                onMouseDown={this.onMouseDown}
-                onMouseEnter={this.onMouseEnter}
-                onMouseMove={this.onMouseMove}
+                onMouseDown={option.value && this.onMouseDown}
+                onMouseEnter={option.value && this.onMouseEnter}
+                onMouseMove={option.value && this.onMouseMove}
```
morethanfire, line 75: Do we need to add this security here?

src/inputs/phoneInput/phoneInputCountryDropdown.jsx
```diff
@@ -55,7 +55,7 @@ class PhoneInputCountryDropdown extends React.PureComponent {
 
         if (prevProps.value !== value && prevState.selectedOption.value !== value) {
             this.setState({
-                selectedOption: _.find(options, (o) => !o.divider && o.value === value),
+                selectedOption: _.find(options, (o) => !o.divider && o.value && o.value === value),
```
morethanfire, line 58: Do we need these changes here?

src/inputs/phoneInput/phoneInputCountryDropdownItem.jsx
raoufr: so the idea here is that if the option does not have a value, then we should not treat it as a selectable option and just ignore events on it. So will be display only item.

src/inputs/phoneInput/phoneInputCountryDropdown.jsx
raoufr: this is just being extra defensive again in the case of not having a value. But you might have a point that at least in theory there shouldn't be any path that would lead us here without a value on the option item. I can remove, let me know.

raoufr:
> Looks the last we need is to not return `international` property.

@morethanfire  I think it is OK to return the `International` item to keep the use of this heading in the drop down, so there will be no change to the content. At the same time we are able now to disable any action upon its selection which also resolves the original bug. What do you think? Let's definitely linger and confirm this in the People scrum. Thanks @morethanfire!

PR #194, opened 2020-08-26
[dev] PBI AB#34262 / Task AB#33950 Updated <DurationPicker> component

groberts314:
Update `<DurationPicker>` component to use Dropdown select menus instead of number inputs to configure the various units of time in the duration

Comps example: https://www.sketch.com/s/cc2a101d-22ab-4e15-99ff-41d29f679094/a/W7vLnK

~Unit Testing is WIP.  Right now I just have a few test cases mounting it with mock theme provider.  I think I have to separately export inner component and do some test cases with `shallow()` to exercise the functionality properly.~ Unit Testing is complete as well.

Screenshot of updated component and documentation page:

image


groberts314:
Unit Test Coverage at 100% for all metrics:

image

docs/src/inputs/durationPicker/durationPicker.jsx
```diff
@@ -1,351 +1,199 @@
...
-                        style={{ height: '175px', padding: '22px' }}
+/* eslint-disable import/no-named-default, import/extensions */
+import { default as durationPickerDoc } from '!!@advclb/react-docgen-loader!react-cm-ui/inputs/durationPicker/durationPicker';
+/* eslint-enable import/no-named-default, import/extensions */
+
+function DocsDurationPicker() {
+    const descriptionCopy = durationPickerDoc.description;
+
+    return (
+        <Main page="duration_picker">
```
morethanfire, line 26: Doesn't have to be a blocker, but I just started a pattern to make the creation of these doc pages a little more simple and one the changes is to use the `durationPickerDoc.name` value here as the value for the `page` prop. 

```jsx
const name = durationPickerDoc.name;

return (
    <Main page={camelCase(name)}>
        ...
    </Main>
);

File: docs/src/inputs/durationPicker/durationPicker.jsx [hunk 2]
```diff
@@ -1,351 +1,199 @@
...
+            <TitleBar title="Duration Picker" />
```
morethanfire, line 27: Same here with the title, we can use the `name` value for the `title` prop's value.

docs/src/inputs/durationPicker/durationPicker.jsx
```diff
@@ -1,351 +1,199 @@
...
+                    rawCode={require('!!raw-loader!./durationPickerDisabledExample').default}
+                >
+                    <DurationPickerDisabledExample />
+                </Example>
+
+                <ComponentApi
+                    docs={[
+                        durationPickerDoc,
+                    ]}
+                />
```
morethanfire, line 193: Please add the ComponentVersionIdentifier here.

src/inputs/durationPicker/__test__/durationPicker.test.js
```diff
@@ -0,0 +1,622 @@
...
+import React from 'react';
+import { BEM_BLOCK_NAME } from '../constants';
+import DurationPicker from '../durationPicker';
+import StyledDurationPicker from '../styledDurationPicker';
+import MockedTheme from '../../../testUtils/mockedTheme';
+
+describe('<DurationPicker />', () => {
+    it('Can render without problems (all default props)', () => {
+        const wrapper = mount(
+            <MockedTheme>
```
morethanfire, line 16: Let's merge my #192 PR and then update mount methods in this file.

src/inputs/durationPicker/durationPicker.jsx
```diff
@@ -1,53 +1,138 @@
...
+     */
     showSeconds: PropTypes.bool,
+    /**
+     * Indicate whether the DurationPicker includes a Years component.
+     */
     showYears: PropTypes.bool,
-    style: PropTypes.shape({}),
+    /**
+     * The value of the DurationPicker.  Should be a `Duration` object from the Moment JS library.
+     */
```
morethanfire, line 116: Nice, these doc changes are going to be so helpful!

src/inputs/durationPicker/index.js
```diff
@@ -1 +1 @@
-export { default } from './durationPicker';
+export { default } from './styledDurationPicker';
```
morethanfire, line 1: Curious as to why we are doing this?

src/versions.js
```diff
@@ -1,7 +1,15 @@
 const versions = {
     'react-cm-ui': {
         package: '9.15.0',
-        components: {},
+        components: {
+            inputs: {
+                durationPicker: {
+                    devLibraryVersion: '2.0.0',
+                    designLibraryVersion: '2.0.0',
```
morethanfire, line 8: I am wondering if we should put a todo for this line too?

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Yep, I'll make it so!

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Will do!

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Will do

src/inputs/durationPicker/__test__/durationPicker.test.js
groberts314: Yep, will do

src/inputs/durationPicker/index.js
groberts314: So, I needed to export the main component itself for some of the unit testing, so I could access state and instance methods ... I couldn't figure how to do that with the mounted and wrapped component.  Maybe there is a way?  Anyhow, so I exported both from the one file, and then  the `react-docgen` package (used to generate the "Component API" documentation) was NOT happy ... see failed validation build https://dev.azure.com/saddlebackchurch/b8a0d0bb-a4fe-48e3-9a60-2f1440145fc4/_build/results?buildId=20429.

So this is what I came up with to appease all of the things.  Kind of strange, but it works.  See my comments on 0a8ea90 where I tried to explain.

If there is a way to do all of my testing using the HOC-wrapped and mounted component, so we don't have to export the inner component, I'm willing to refactor a bit.  I might need a bit of help with that.

Lemme know your preferences.

src/versions.js
groberts314: Would you prefer an empty string and a `// TODO: Add this when it is available` comment?  I'm okay doing it.  Lemme know.

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Hmm ... I don't think that is `dev` branch yet  ... It looks to be brand spanking new in `feature/10.0.0--grid-2.0.0` / #191 ... right?

groberts314: Well, it is also in `feature/doc_page_components` / #192 ... we definitely need to merge that guy first it looks like.

src/versions.js
groberts314: Changed it to empty string and `// TODO` comment in ed7cd01.

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Fixed in ed7cd01.
For some reason it was `displayName` not `name`.  I wonder if that is maybe a difference between class components and function components.

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Fixed in ed7cd01.
For some reason it was `displayName` not `name`.  I wonder if that is maybe a difference between class components and function components.

docs/src/inputs/durationPicker/durationPicker.jsx
groberts314: Fixed in 629045a.

image

src/inputs/durationPicker/__test__/durationPicker.test.js
groberts314: Fixed in 629045a.

groberts314:
@morethanfire I believe everything has been resolved.  Please have another go at it, at your earliest convenience.

Let me know if you wanna follow up on [this concern](https://github.com/saddlebackdev/react-cm-ui/pull/194#discussion_r477575384) further, or if we wanna take it as-is.

Thank you sir!

docs/src/inputs/durationPicker/durationPicker.jsx
morethanfire: Yes, I saw that accident in my branch as well. Thanks for fixing!

docs/src/global/componentVersionIdentifier.jsx
```diff
@@ -21,7 +21,13 @@ function ComponentVersionIdentifier(props) {
...
     } = props;
 
-    const pathNameArray = filter(pathname.split('/'), (segment) => segment);
+    const pathNameArray = map(
+        filter(
+            pathname.split('/'),
+            (segment) => !!segment,
+        ),
+        (segment) => segment.replace(/-([a-z])/g, (match, capture) => (capture.toUpperCase())), // e.g. 'duration-picker' => 'durationPicker'
+    );
```
morethanfire, line 30: 👍

src/inputs/durationPicker/index.js
morethanfire: I am not sure I am understanding, `wrapper.find('FooComponent').instance()` doesn't work with the `withStyles` HOC?

groberts314: I can try it again.  Specifically I recall having trouble using `.setProps()` in the test cases that are testing the logic in `componentDidUpdate()` ... I was getting an error message with something to the effect of only being able to call that from the root node.

groberts314: @morethanfire Can we maybe spend a few minutes discussing this issue?
Basically, I want to arrive at a decision of whether this is acceptable as-is, or if I need to make an effort to refactor the unit tests to avoid having to do this.  It'd be nice to get this resolve today if possible, obviously given other priorities and workload.

groberts314: Dude, you were totally right.  All fixed up in 058389b.  Sorry 'bout that.  I've learned now.

morethanfire: Oh, Nice! Thank you.

PR #206, opened 2020-09-16
[dev] No PBI - Fixed Input Outlines

morethanfire:


src/dataDisplay/icon/icon.jsx
```diff
@@ -90,6 +91,10 @@ class Icon extends React.PureComponent {
         }
     }
 
+    onMouseDown(event) {
+        event.preventDefault();
+    }
```
groberts314, line 96: Is this _always_ the right thing to do?  There are no exceptions, where this is not correct?

src/inputs/button/button.jsx
```diff
@@ -300,6 +305,10 @@ class Button extends React.PureComponent {
         }
     }
 
+    onMouseDown(event) {
+        event.preventDefault();
+    }
```
groberts314, line 310: Same question here.

groberts314 (approved):
Looks pretty good.  I just left some questions seeking confirmation that it is always correct to prevent default on mouse down events.  If you're pretty sure, then I'm good with it and :shipit:

src/inputs/button/button.jsx
morethanfire: I found an answer in a GitHub thread somewhere that doing this will block the client from bubbling up the event to focus the input. I don't believe this is the best answer to the problem, but I believe design needs to design the different states. For now, it works in Chrome, Safari, and Firefox: when tabbing the input is outlined and when clicking input it is not outlined.

PR #211, opened 2020-09-25
[dev] Bug AB#33296: [Person Add 2.0] - Child/Student Records in duplicate list is displaying incorrect age

raoufr:


sircris (approved):
Seems fairly simple. Do we need to handle various formats of the date (e.g. long, Date, etc)? I'm not asking for the immediacy but for the future.

War777 (approved):
Looks pretty simple.

morethanfire (approved):
:shipit:

PR #225, opened 2020-10-17
[dev] Bug AB#33296: [Person Add 2.0] - Child/Student Records in duplicate list is displaying incorrect age

raoufr:
…list is displaying incorrect age


morethanfire (approved):
Yep, this makes sense. :shipit:

PR #239, opened 2020-11-13
[feature/cululative--10.0.0] AB#33496 - Removed Loader Component

cbrewer1689:


groberts314 (approved):
Looks good.  Is gonna be _a lot_ of work to update HC to replace all the `<Loader />`s with `<ActivityIndicator />`s.

cbrewer1689:
So true.

PR #245, opened 2020-11-17
[feature/cumulative--10.0.0] AB#33499 - Banner Component

fantaJinMode:


src/dataDisplay/banner/banner.jsx
```diff
@@ -30,19 +56,50 @@ const propTypes = {
         'teal',
         'warning',
     ]),
```
groberts314, line 58: Dumb question: Why are colors mixed into a thing called `level`?
**_Not a Blocker_** since this is how it's been (likely for some time).  But, since I'm reviewing, I'm calling it out, 'cause it's weird.

groberts314 (approved):
Looks pretty good to me.

src/dataDisplay/banner/banner.jsx
fantaJinMode: @groberts314 It was previously implemented like this. But can be a good opportunity to rename it. I found some code below like this
```
const containerClasses = ClassNames('ui', 'banner', className, {
            'banner-level-error': level === 'error',
            'banner-level-purple': level === 'purple', // Rename when better defined.
            'banner-level-secondary': !level || level === 'secondary',
            'banner-level-success': level === 'success',
            'banner-level-teal': level === 'teal', // Rename when better defined.
            'banner-level-warning': level === 'warning',
            'banner-type-alert': type === 'alert',
            'banner-type-notification': type === 'notification',
        });
```

groberts314: Yeah, I'm thinking since this is a brand new, shiny version for 10.0.0, let's do this right.

@cbrewer1689 Any thoughts here about how accommodate some of the color schemes in with the notion of "level" (which imply a color)?

fantaJinMode: @groberts314 I have introduced a new prop `color`. Could you please have a look once ?

cbrewer1689 (approved):
Looks good.

PR #250, opened 2020-11-30
[dev] No PBI - Spread styles object for each columns on Data Card.

alejosod:


cbrewer1689 (approved):
Looks good.

src/dataDisplay/dataCards/dataCardColumn.jsx
```diff
@@ -37,6 +43,7 @@ function DataCardColumn(props) {
                 style={{
                     marginBottom: column.width ? '11px' : null,
                     width: column.width,
+                    ...column.style,
```
groberts314, line 46: Do we need to defensively check for `style`?
Maybe `...(style ?? {})` or something.

groberts314 (changes requested):
One concern about being defensive if `column.style` is nullish.  Otherwise looks good.

src/dataDisplay/dataCards/dataCardColumn.jsx
cbrewer1689: Yeah, I thin we need to be defensive. Good catch.

cbrewer1689: ...

groberts314: Looks better now.  Resolving.

groberts314 (approved):

War777 (approved):
Looks good to me

PR #257, opened 2020-12-11
[dev] AB#40135 - Removed Grid Container In ActionsBar If `columns` Empty.

cbrewer1689:


src/surfaces/actionBar/actionBar.jsx
```diff
@@ -229,12 +236,12 @@ class ActionBar extends React.PureComponent {
                                                 </DropdownButton>
                                             )}
 
-                                            {list && _.isArray(list) && (
+                                            {list && isArray(list) && (
```
groberts314, line 239: `list && isArray(list)` is probably redundant.  Could probably just be `isArray(list)`

src/surfaces/actionBar/actionBar.jsx
```diff
@@ -245,7 +252,7 @@ class ActionBar extends React.PureComponent {
                                                             iconTable,
                                                         } = item;
 
-                                                        const divide = _.isUndefined(item.divide) ||
+                                                        const divide = isUndefined(item.divide) ||
                                                             item.divide;
```
groberts314, line 256: Maybe Not a Blocker because prior code was like this, but this is weird.  So, if `item.divide` is undefined, `divide` will be `true`; otherwise, it is whatever it is.  Is that really the intent?

Also, this just confuses matters further (it's a Boolean, but it defaults to `null`):
https://github.com/saddlebackdev/react-cm-ui/blob/3eed68f2a07caf3a78ec63d3dd84c73d6bc4fbfa/src/dataDisplay/list/listItem.jsx#L7-L21

groberts314 (changes requested):
Looks okay.  I have one small requested simplification, and one question.

src/surfaces/actionBar/actionBar.jsx
cbrewer1689: Fixed.

src/surfaces/actionBar/actionBar.jsx
cbrewer1689: Fixed.

src/surfaces/actionBar/actionBar.jsx
groberts314: Looks good.  Resolving.

src/surfaces/actionBar/actionBar.jsx
groberts314: Yeah, seems cleaner as of f47470e.  Thanks for fixing that.  Resolving.

groberts314 (approved):

PR #262, opened 2020-12-17
[dev] PBI AB#36142 Update Events Central with Mobile Components - Overview Page

IlyaRadinsky:
This PR should align the functionality of the Filter's nested toggles with `Dropdown`:
- Added `singleSelection` feature;
- Added `clearable` feature (to be able to create the single selection switch)
- Allowing to use just `value` data for selection;

Also improved Filter's `multiSelect` option:
- Allowing to use just `value` data for selection;

### DIFF
--- a/src/surfaces/filtersDrawer/filtersDrawer.jsx
+++ b/src/surfaces/filtersDrawer/filtersDrawer.jsx
@@ -120,6 +120,7 @@ const propTypes = {
                         PropTypes.string,
                     ]),
                 })),
+                singleSelection: PropTypes.bool,
                 value: PropTypes.arrayOf(PropTypes.shape({
                     label: PropTypes.string,
                     value: PropTypes.oneOfType([
@@ -227,7 +228,9 @@ class FiltersDrawer extends React.Component {
         const { nestedTogglesData } = this.state;
         let selectedOptions;
 
-        if (some(nestedTogglesData.value, selectedOption)) { // Subtract
+        if (nestedTogglesData.singleSelection) {
+            selectedOptions = [selectedOption];
+        } else if (some(nestedTogglesData.value, selectedOption)) { // Subtract
             selectedOptions = filter(nestedTogglesData.value, (d) => (
                 d.value !== selectedOption.value
             ));
--- a/src/surfaces/filtersDrawer/filtersDrawer.jsx
+++ b/src/surfaces/filtersDrawer/filtersDrawer.jsx
@@ -1,9 +1,11 @@
 import {
     cloneDeep,
     differenceBy,
+    find,
     filter,
     isArray,
     isEmpty,
+    isObject,
     map,
     mapValues,
     some,
@@ -230,7 +232,7 @@ class FiltersDrawer extends React.Component {
                 d.value !== selectedOption.value
             ));
         } else { // Add
-            selectedOptions = sortBy([...nestedTogglesData.value, selectedOption], ['value']);
+            selectedOptions = sortBy([...(nestedTogglesData.value) || [], selectedOption], ['value']);
         }
 
         const newNestedTogglesData = mapValues(nestedTogglesData, (data, key) => {
@@ -323,7 +325,10 @@ class FiltersDrawer extends React.Component {
                                             nestedTogglesOptionLabelKeyNum += 1;
                                             const isSelected = some(
                                                 nestedTogglesData.value,
-                                                option,
+                                                (v) => {
+                                                    const selectedValue = isObject(v) ? v.value : v;
+                                                    return option.value === selectedValue;
+                                                },
                                             );
 
                                             return (
@@ -626,8 +631,10 @@ class FiltersDrawer extends React.Component {
                                                             <div>
                                                                 {map(
                                                                     nestedToggles.value,
-                                                                    (option) => {
+                                                                    (value) => {
                                                                         nestedTogglesValueLabelKeyNum += 1;
+                                                                        const selectedValue = isObject(value) ? value.value : value;
+                                                                        const option = find(nestedToggles.options, { value: selectedValue });
 
                                                                         return (
                                                                             <FiltersDrawerNestedTogglesValueLabel
--- a/src/surfaces/filtersDrawer/filtersDrawer.jsx
+++ b/src/surfaces/filtersDrawer/filtersDrawer.jsx
@@ -1,6 +1,6 @@
 import {
     cloneDeep,
-    differenceBy,
+    differenceWith,
     find,
     filter,
     isArray,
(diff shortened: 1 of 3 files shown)


src/surfaces/filtersDrawer/filtersDrawer.jsx
```diff
@@ -225,12 +229,14 @@ class FiltersDrawer extends React.Component {
...
-        if (some(nestedTogglesData.value, selectedOption)) { // Subtract
+        if (nestedTogglesData.singleSelection) {
+            selectedOptions = [selectedOption];
+        } else if (some(nestedTogglesData.value, selectedOption)) { // Subtract
             selectedOptions = filter(nestedTogglesData.value, (d) => (
                 d.value !== selectedOption.value
             ));
         } else { // Add
-            selectedOptions = sortBy([...nestedTogglesData.value, selectedOption], ['value']);
+            selectedOptions = sortBy([...(nestedTogglesData.value) || [], selectedOption], ['value']);
```
groberts314, line 239: Are the parentheses necessary here?
Or was the intent `...(nestedTogglesData.value || [])`?

src/surfaces/filtersDrawer/filtersDrawer.jsx
```diff
@@ -323,7 +329,10 @@ class FiltersDrawer extends React.Component {
                                             nestedTogglesOptionLabelKeyNum += 1;
                                             const isSelected = some(
                                                 nestedTogglesData.value,
-                                                option,
+                                                (v) => {
+                                                    const selectedValue = isObject(v) ? v.value : v;
+                                                    return option.value === selectedValue;
```
groberts314, line 334: Is there any way we can avoid having to "sniff" if we're a primitive or an option shaped object (`{ label, value }`) here, as well as down on Lines 598 and 644?  Was there a particular use case you ran into that really is different than other use cases using this component so far?  Is this what you mean by "allowing to use just `value` data for selection"?

I'm not necessarily opposed, but I want to ensure that this is the best way, before we complicate the code.

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
```diff
@@ -1,10 +1,13 @@
+import { get, noop, size } from 'lodash';
```
groberts314, line 1: Up to @cbrewer1689 as the main steward of this codebase as to whether this is Blocker or NAB (and maybe @War777 can weigh in as well) but I think we tend to prefer one thing per line these days:
```JavaScript
import {
    get,
    noop,
    size,
} from 'lodash';
```

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
```diff
@@ -25,8 +28,10 @@ class FiltersDrawerNestedTogglesValueLabel extends React.PureComponent {
     }
 
     render() {
-        const { option: { label } } = this.props;
+        const { nestedTogglesData, option: { label } } = this.props;
         const containerClasses = ClassNames('page_filters_drawer--nested_toggles_value_label');
+        const clearable = get(nestedTogglesData, 'clearable', true);
+        const canClear = clearable || size(nestedTogglesData.value) > 1;
```
groberts314, line 34: Really OR here?  Not AND?
i.e. not this?
```JavaScript
const canClear = clearable && size(nestedTogglesData.value) > 1;
```

Also, what if the size is only 1?  There's no circumstance under which that'd be clearable?  i.e. no use case for "nothing selected" versus "the one thing selected"?  I guess that'd be weird ... we shouldn't be using nested toggles in that case; we should be using a checkbox.  But just confirming the intent here, and whether we've thought through all conceivable use cases.

groberts314 (changes requested):
Looks pretty good, but I do have some questions and some possible requested changes.

groberts314:
**[Product Backlog Item 36142](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/36142) | Update Events Central with Mobile Components - Overview Page** is slated for 1.33.0 release (**vNextNext** not **vNext** ... second major release of 2021) so we do have some time.  Although this might be blocking @IlyaRadinsky 's development work in Healthy Church.

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
IlyaRadinsky: @groberts314  `clearable` means here the ability to remove all options (like `Dropdown` same prop), so OR should be here.

src/surfaces/filtersDrawer/filtersDrawer.jsx
IlyaRadinsky: @groberts314 the component will crash if it has no value (`null` or `undefined`), it's helpful update.

src/surfaces/filtersDrawer/filtersDrawer.jsx
IlyaRadinsky: @groberts314 you are right - at the moment you should "sniff" option shaped object for correct initialization, and I'm really sure that this a helpful update. I had the Dropdown similar functionality in mind.

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
IlyaRadinsky: Fixed.

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
groberts314: Just double-checking.  I'll take your word for it! :-)

src/surfaces/filtersDrawer/filtersDrawer.jsx
groberts314: Sorry; I was asking if it could be simplified to this:
```JavaScript
selectedOptions = sortBy([...nestedTogglesData.value || [], selectedOption], ['value']);
```
(no parentheses around `nestedTogglesData.value` because they appear unnecessary, unless I'm missing something...)

src/surfaces/filtersDrawer/filtersDrawerNestedTogglesValueLabel.jsx
groberts314: Thank you good sir!  Resolving.

cbrewer1689 (commented):
Seems good to me once @groberts314 requests are satisfied.

groberts314:
I'm thinking I just want to follow up on [this thread](https://github.com/saddlebackdev/react-cm-ui/pull/262#discussion_r545021056) and then I'm good to go.

src/surfaces/filtersDrawer/filtersDrawer.jsx
IlyaRadinsky: Fixed

groberts314: Thank you sir for the fix!  Apologies if the initial observation here was unclear, causing a lot of much back-and-forth.
Resolving.

groberts314 (approved):

groberts314:
:shipit:

PR #270, opened 2021-02-01
[dev] AB#41363 - Button Dropdown Active State

cbrewer1689:


src/colors/cyan.js
```diff
@@ -1,6 +1,6 @@
 const cyan = {
     500: '#00aeef',
-    600: '#0093c6',
+    600: '#1c93c4',
```
groberts314, line 3: This is a sweeping change, right?  Just confirming it is what is intended?

groberts314 (approved):
LGTM

src/colors/cyan.js
cbrewer1689: Yes, this was intended to match what design has in the docs.

groberts314: Plus, not much of a difference

cyan_600_switcheroo

PR #278, opened 2021-04-14
[dev] BUG AB#45246 [Events Central][Mobile] - Tapping on [ x ] icon next to the campus name does not remove it from the filters

IlyaRadinsky:
Fixes filtering the options in case when an option item is not an object.


groberts314:
Note: This is needed for AB#45246 which currently has Proposed Release Version 1.34.0.
So that will help us assign a milestone / React CM UI release.

image

groberts314 (approved):
Looks good to me.

cbrewer1689 (approved):
@groberts314, how do you feel about this PR potentially being a 9.24.0 which could potentially be in HC's 1.33.0 release? I say this because of the PersonCoreMilestone PR updates. I am not sure whether we are going to take that or not for 1.33.0. I am just playing things by ear.

groberts314:
Just for the record, it was agreed not to include this in the same release as #279 (which is required for HC 1.33.0), simply out of an overabundance of caution on regression surface area for the already stressed HC 1.33.0 release.

Thanks for updating the milestones accordingly @cbrewer1689 !

PR #305, opened 2021-06-15

cbrewer1689:


groberts314 (approved):
Looks good.
Are we needing to be defensive due to issues that arise in unit testing?

cbrewer1689:
Yes, that is correct. Unit testing is causing us more work. But, honestly, I think it is okay.

PR #307, opened 2021-06-22
[release/10.0.3-hotfix] AB#47299: [Events > Check-in] [UI] - Person Relationship records are Not Scrolling horizontally & Broken display of Person record After you adjust Zoom

cbrewer1689:


sircris (approved):
So essentially, you moved the calling of the onCloseComplete method within callback of the setState call in the onClose() method? Does this yield the intended results?

cbrewer1689:
Yep. We are going to do this to be sure that everything has been cleaned up before calling `onCloseComplete`.

groberts314 (approved):
LGTM

PR #313, opened 2021-07-06
[dev] Typography Updates

cbrewer1689:


groberts314 (approved):
LGTM

We should consider removing `/components/data-display/typography` route and stubbed components from the Docs site.

groberts314:
> We should consider removing `/components/data-display/typography` route and stubbed components from the Docs site.

Don't mind me! =)  If you have legit future plans for that guy we can keep it around.

PR #315, opened 2021-07-10

cbrewer1689:
In this PR:
- Fixed the dropdown deprecated docs page
- Added a bunch of new simple Select options for our examples.
- Removed `autoScrollSelection`. No need for this. it should have been doing this all along without an extra prop. It now scrolls to the selected option on open by default.
- Fixed disabled option styling.
- Fixed scrolling to focused option, sense our component is using its own menu scroller.
- Added and updated option props

### DIFF
--- a/docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
+++ b/docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
@@ -1285,94 +1285,90 @@ class ModulesDropdown extends React.Component {
                         </Header>
 
                         <Grid>
-                            <Grid.Row stackable>
-                                <Grid.Column width={6} tablet={4} laptop={2}>
-                                    <Dropdown button buttonColor="alert" collapseMenuOnChange text="Alert" style={{ maxWidth: '200px', width: '100%' }}>
+                            <Grid.Column width={6} tablet={4} laptop={2}>
+                                <Dropdown button buttonColor="alert" collapseMenuOnChange text="Alert" style={{ maxWidth: '200px', width: '100%' }}>
+                                    <Dropdown.Item label="Option 1" />
+                                    <Dropdown.Item label="Option 2" />
+                                </Dropdown>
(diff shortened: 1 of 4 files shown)


groberts314 (approved):
LGTM

docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
```diff
@@ -1285,94 +1285,90 @@ class ModulesDropdown extends React.Component {
                         </Header>
 
                         <Grid>
-                            <Grid.Row stackable>
-                                <Grid.Column width={6} tablet={4} laptop={2}>
-                                    <Dropdown button buttonColor="alert" collapseMenuOnChange text="Alert" style={{ maxWidth: '200px', width: '100%' }}>
+                            <Grid.Column width={6} tablet={4} laptop={2}>
```
War777, line 1288: We're missing the spacing between the Dropdown examples, is that expected?

image

War777 (approved):
Looks good, I'm only wondering about a little question about spacing in the Dropdown Deprecated Button Docs. Thanks @cbrewer1689

docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
cbrewer1689: Yeah, thanks for checking. Yeah, I think i am okay with that out of time and because it's deprecated.

War777: No worries then, I agree with you, thanks.

fantaJinMode (approved):
LGTM 👍

docs/src/inputs/dropdownDeprecated/dropdownDeprecated.jsx
groberts314: Is it really hard to give them a little bit of margin?

PR #316, opened 2021-07-12

cbrewer1689:
In this PR:
- KeyDown Events for No and Yes buttons.
- Focuses No button after child element click or enter key.

### DIFF
--- a/src/inputs/prompt/prompt.jsx
+++ b/src/inputs/prompt/prompt.jsx
@@ -1,7 +1,6 @@
 import {
     isFunction,
     isUndefined,
-    noop,
 } from 'lodash';
 import ClassNames from 'classnames';
 import PropTypes from 'prop-types';
--- a/src/inputs/prompt/prompt.jsx
+++ b/src/inputs/prompt/prompt.jsx
@@ -1,6 +1,7 @@
 import {
     isFunction,
     isUndefined,
+    noop,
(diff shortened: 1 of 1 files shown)


src/inputs/prompt/prompt.jsx
```diff
@@ -155,30 +162,49 @@ class Prompt extends React.Component {
...
-        if (show) { return false; }
+        if (show) { return; }
 
         if (!isUndefined(onClick)) {
             onClick(option);
         } else {
             this.setState({ show: true });
         }
-
-        return false;
```
groberts314, line 178: We're sure `return false;` is unnecessary here?  It wasn't like serving to stop the event from propagating or anything like that, right?  I trust you ... just triple-checking ;-)

groberts314 (approved):
LGTM

fantaJinMode (approved):
LGTM 👍

src/inputs/prompt/prompt.jsx
cbrewer1689: It was to appease eslint because the a few lines above, we were returning `false`. I don't see why we'd need to return anything other than to get out of there if `show` is `true`

groberts314: Sounds good.  Resolving.

PR #317, opened 2021-07-16
[dev] PBI AB#48476: Select `noResultsMessage` prop

cbrewer1689:


src/inputs/select/select.jsx
```diff
@@ -79,6 +79,11 @@ const propTypes = {
      * A Select can have multiple values
      */
     multiple: PropTypes.bool,
+    /**
+     * Placeholder displayed when there are no matching search results or a falsy value to hide it
+     * (can also be a react component)
+     */
+    noResultsText: PropTypes.string,
```
groberts314, line 86: > or a falsy value to hide it
> (can also be a react component)

Do we maybe need `PropTypes.oneOfType[]` here with some different options?

groberts314 (approved):
Looks good, but one question about the prop type.

src/inputs/select/select.jsx
cbrewer1689: Yeah, you are right. Their docs were wrong.

cbrewer1689: Fixed

groberts314: So that will allow a component or a string.  By "a falsy value to hide it" I presume we mean an empty string? ;-)

groberts314: Should we fork 'em and submit a PR to fix their docs? 😁

cbrewer1689: Hahaha! Actually it's a good idea, but I am sure because we're on v1.3.0 they going to say "get that corn out of my face."

PR #324, opened 2021-08-20
[dev] PBI AB#51574: Palette Update to support the DX-React-Scheduler-Material-UI package

cbrewer1689:


src/styles/createPalette.js
```diff
@@ -133,13 +134,13 @@ export default function createPalette() {
         return contrastText;
     };
 
-    const mode = 'light';
-    const modes = { dark, light };
+    const type = 'light';
+    const types = { dark, light };
```
groberts314, line 138: Just curious: Why "type" versus "mode"?

groberts314 (approved):
LGTM.  Just one question about `mode` => `type` rename.

src/styles/createPalette.js
cbrewer1689: Very good question. `mode` is from MUI's next, `v5.x.x` release. We're using the latest `v4.x.x` release. So apparently when I was updating this file a few months ago, when doing implementing the Snackbar, I didn't change the version number in MUI's GitHub repo and grabbed the wrong code. This fixes the issue so that things like DX-React-Scheduler-Material-Ui (uses MUI v4) to not cause an error in our app.

PR #325, opened 2021-08-21
[dev] AB#51574: Button Upgrade

cbrewer1689:

PR #327, opened 2021-09-01
[dev] AB#51574: List and Divider Updates

cbrewer1689:


groberts314:
@cbrewer1689 I'll take a look as soon as I can.

FYI @fantaJinMode may require a small update for a forthcoming React CM UI release for Connection Forms 2.0 (for HC 1.39.0).  We need to make the Select button on the Person 2.0 Panel optional, so it can be excluded from the placement here (see left-hand side of the main content area): https://www.sketch.com/s/877c6c65-893c-456e-9d33-9bd002036677/a/YGvdgmM

Is 10.2.0 the right release version for that to target as well?

cbrewer1689:
>FYI @fantaJinMode may require a small update for a forthcoming React CM UI release for Connection Forms 2.0 (for HC 1.39.0). We need to make the Select button on the Person 2.0 Panel optional, so it can be excluded from the placement here (see left-hand side of the main content area): https://www.sketch.com/s/877c6c65-893c-456e-9d33-9bd002036677/a/YGvdgmM
>
> Is 10.2.0 the right release version for that to target as well?

Got it. With how we're doing v10 release, I believe it should go into a patch release. It however can go into a minor release, such a `10.2.0`, if the timing is right. No sense of making the releases too complicated, imho.

cbrewer1689:
Do not merge until PR #325 is merged.

PR #329, opened 2021-09-03
[dev] AB#25926 - [W.2] - Ability to create an entry for a connection form - Modification of default props on person panel react-cm-ui component 

fantaJinMode:
PR contains:
- Modification of default props for person panel component
- If `personId` isn't passed for `summaryData` then don't display personId

#note: @War777 I have been reverting some code which I believe was changed by you. Could you please verify it once?

### DIFF
--- a/src/dataDisplay/personPanel/personPanelDetails.jsx
+++ b/src/dataDisplay/personPanel/personPanelDetails.jsx
@@ -200,7 +200,10 @@ const defaultProps = {
     isExpanded: false,
     isMobile: false,
     otherDataGroups: null,
-    selectButtonProps: {},
+    selectButtonProps: {
+        label: 'Select',
+        disable: false,
+    },
     showSelectButton: true,
     viewRecordButtonProps: {},
 };
--- a/src/dataDisplay/personPanel/personPanelDetails.jsx
+++ b/src/dataDisplay/personPanel/personPanelDetails.jsx
@@ -173,6 +173,10 @@ const propTypes = {
         promptId: PropTypes.string,
     }),
     /**
+     * If `false`, select button isn't displayed for PersonPanelDetails.
+     */
+    showSelectButton: PropTypes.bool,
+    /**
      * Button `props` to setup the View Record button.
      */
     viewRecordButtonProps: PropTypes.shape({
@@ -197,6 +201,7 @@ const defaultProps = {
     isMobile: false,
     otherDataGroups: null,
     selectButtonProps: {},
+    showSelectButton: true,
     viewRecordButtonProps: {},
 };
 
@@ -664,6 +669,7 @@ function PersonPanelDetails(props) {
         isMobile,
         otherDataGroups,
         selectButtonProps,
+        showSelectButton,
         viewRecordButtonProps,
     } = props;
 
@@ -888,28 +894,31 @@ function PersonPanelDetails(props) {
 
                     {children}
 
-                    {(!isEmpty(selectButtonProps) || !isEmpty(viewRecordButtonProps)) && (
-                        <div
-                            className={classes.actions}
-                        >
-                            {!isEmpty(selectButtonProps) && (
-                                <PersonPanelDetailsActionButton
-                                    label="Select"
-                                    // eslint-disable-next-line react/jsx-props-no-spreading
-                                    {...selectButtonProps}
-                                />
-                            )}
+                    {
+                        ((!isEmpty(selectButtonProps) && showSelectButton) ||
+                        !isEmpty(viewRecordButtonProps)) && (
+                            <div
+                                className={classes.actions}
+                            >
+                                {!isEmpty(selectButtonProps) && showSelectButton && (
+                                    <PersonPanelDetailsActionButton
+                                        label="Select"
+                                        // eslint-disable-next-line react/jsx-props-no-spreading
+                                        {...selectButtonProps}
+                                    />
+                                )}
 
-                            {!isEmpty(viewRecordButtonProps) && (
-                                <PersonPanelDetailsActionButton
-                                    // eslint-disable-next-line react/jsx-props-no-spreading
-                                    {...viewRecordButtonProps}
-                                    label="View Record"
-                                    outlined
-                                />
-                            )}
-                        </div>
-                    )}
+                                {!isEmpty(viewRecordButtonProps) && (
+                                    <PersonPanelDetailsActionButton
+                                        // eslint-disable-next-line react/jsx-props-no-spreading
+                                        {...viewRecordButtonProps}
+                                        label="View Record"
+                                        outlined
+                                    />
+                                )}
+                            </div>
+                        )
+                    }
                 </div>
             </Collapse>
         </div>
--- a/src/dataDisplay/personPanel/personPanelDetails.jsx
+++ b/src/dataDisplay/personPanel/personPanelDetails.jsx
@@ -196,10 +196,7 @@ const defaultProps = {
     isExpanded: false,
     isMobile: false,
     otherDataGroups: null,
-    selectButtonProps: {
-        label: 'Select',
-        disable: false,
-    },
+    selectButtonProps: {},
     viewRecordButtonProps: {},
 };
 
--- a/src/dataDisplay/personPanel/personPanelDetails.jsx
+++ b/src/dataDisplay/personPanel/personPanelDetails.jsx
@@ -836,6 +836,8 @@ function PersonPanelDetails(props) {
     );
 
     const shouldDataGroupsRender = !isEmpty(data) && !isEmpty(dataGroupsColumns);
+    const shouldRenderSelectButton = showSelectButton && !isEmpty(selectButtonProps);
+    const shouldRenderViewButton = !isEmpty(viewRecordButtonProps);
 
     return (
         <div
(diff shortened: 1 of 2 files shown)


src/dataDisplay/personPanel/personPanelDetails.jsx
```diff
@@ -196,10 +196,7 @@ const defaultProps = {
     isExpanded: false,
     isMobile: false,
     otherDataGroups: null,
-    selectButtonProps: {
-        label: 'Select',
-        disable: false,
-    },
+    selectButtonProps: {},
```
groberts314, line 199: Is this change necessary?  I think we should keep it if possible to avoid regressions.  i.e. by default this button should show and have these props.  And someone (like us) that wishes to suppress this button should have to do something.

Maybe introduce a new `showSelectButton` prop (Boolean, optional, defaults to `true`, and we would set it to `false`).

src/dataDisplay/personPanel/personPanelSummary.jsx
```diff
@@ -577,17 +577,19 @@ function PersonPanelSummary(props) {
...
-                    </Typography>
-                </Grid.Column>
+                        <Typography
+                            className={classes.personId}
+                            variant="caption"
+                        >
+                            {`Id: ${personId}`}
+                        </Typography>
+                    </Grid.Column>
+                )}
```
groberts314, line 592: This is not a bad enhancement, although we might double check if omission of Person ID on our comps (e.g. https://www.sketch.com/s/877c6c65-893c-456e-9d33-9bd002036677/a/Vrzbyeo ) is actually intentional.  We might want to show the Person ID, as this is helpful to our end-users I believe.  Let's double check with Sandra and Hannah at scrum today.

groberts314 (changes requested):
I'm not quite sure the Select button stuff is quite what we should be doing.

In my opinion, let's not mess with the default value for `selectButtonProps`, and let's introduce a new prop `showSelectButton` (Boolean, optional, defaults to `true`, so we can then do `showSelectButton={false}` to suppress the Select button.

src/dataDisplay/personPanel/personPanelSummary.jsx
fantaJinMode: This would also avoid situations like if personId isn't passed then it would show on UI `personId: undefined`

src/dataDisplay/personPanel/personPanelDetails.jsx
fantaJinMode: @groberts314 Make sense if it's to avoid possible regressions, but it can be handled from UI as well. Should I add boolean `showSelectButton` and make those changes? Guys, what are your views? @War777 @cbrewer1689

cbrewer1689: I believe this change seems good. Having the default be `{}` makes sense here, because below, in the render, the label's value is being set to "Select" by default and `disable` is already be set to `false` inside of the PersonPanelDetailsActionButton component.

src/dataDisplay/personPanel/personPanelSummary.jsx
cbrewer1689: If requirements say it is good, I am fine with this change.

cbrewer1689 (approved):

cbrewer1689 (changes requested):
Mmm, thinking about it further, none of my thoughts have changed about the changes below, but I am asking for a requested change here to not rely on an empty selectButtonProps object to not render the button. I think we should have an extra flag, maybe `showSelectButton`.

src/dataDisplay/personPanel/personPanelDetails.jsx
groberts314: So, it sounds like this is fine (change to default value for `selectButtonProps`), but let's also add `showSelectButton` prop as described to govern whether or not to render the button.

^^^ @fantaJinMode

src/dataDisplay/personPanel/personPanelSummary.jsx
groberts314: Yeah, we brought this up during Team S.H.I.E.L.D. scrum this morning with Garrett and Hannah.  This is a good change for this component, but I do think we might bring back Person ID for our particular usage (and have it wrap to the next line).  You know how our end-users love their Person IDs! :-P

src/dataDisplay/personPanel/personPanelDetails.jsx
fantaJinMode: If I am thinking correctly `showSelectButton` prop for `<PersonPanelDetails>` component, Yes ??
If that's the case, should I add another prop `showViewRecordButton` for another button as well ? @groberts314 @cbrewer1689

fantaJinMode: `showSelectButton` Should be prop for <PersonPanelDetails> ??
It would be something like this 
```
selectButtonProps: {
      showSelectButton: false,
  },
```
@groberts314 @cbrewer1689

groberts314: @fantaJinMode I'd say yes.  Please default the props to `true` so that by default nothing changes, and people can suppress the buttons by setting new props to `false`.

fantaJinMode: @groberts314 @cbrewer1689 Fixed. Please check once

groberts314: @fantaJinMode @cbrewer1689 Should we restore the prior default value for `selectButtonProps` or should we leave it as an empty object?

groberts314: I think if we're checking `!isEmpty(selectButtonProps)` below we want to restore the default value so that by default, the Select button is displayed, to avoid regressions.  Does that sound right?  Or am I missing something?

src/dataDisplay/personPanel/personPanelDetails.jsx
```diff
@@ -891,28 +894,31 @@ function PersonPanelDetails(props) {
...
-                                    {...viewRecordButtonProps}
-                                    label="View Record"
-                                    outlined
-                                />
-                            )}
-                        </div>
-                    )}
+                    {
+                        ((!isEmpty(selectButtonProps) && showSelectButton) ||
+                        !isEmpty(viewRecordButtonProps)) && (
```
groberts314, line 899: Can we move some of these expressions out to variables calculated before the `return ( ... );`?
e.g.
```JavaScript
const shouldRenderSelectButton = showSelectButton && !isEmpty(selectButtonProps);
const shouldRenderViewButton = !isEmpty(viewRecordButtonProps);

// ...

return (
    // ...
    {(shouldRenderSelectButton || shouldRenderViewButton) && (
        <div
            className={classes.actions}
        >
             {shouldRenderSelectButton && (
                <PersonPanelDetailsActionButton
                    label="Select"
                    // eslint-disable-next-line react/jsx-props-no-spreading
                    {...selectButtonProps}
                />
             )}
             
             {shouldRenderViewButton && (
                <PersonPanelDetailsActionButton
                    // eslint-disable-next-line react/jsx-props-no-spreading
                    {...viewRecordButtonProps}
                    label="View Record"
                    outlined
                />
             )}
        </div>
    )}
);
```

groberts314:
@fantaJinMode @cbrewer1689 I'm concerned about changing the default value of `selectButtonProps` causing a regression, and I also have an additional suggestion to help with the conditional button rendering logic to avoid computing certain conditions multiple times.  So my status is still at 'request changes'.

src/dataDisplay/personPanel/personPanelDetails.jsx
fantaJinMode: Fixed

fantaJinMode:
> @fantaJinMode @cbrewer1689 I'm concerned about changing the default value of `selectButtonProps` causing a regression, and I also have an additional suggestion to help with the conditional button rendering logic to avoid computing certain conditions multiple times. So my status is still at 'request changes'.

On default logic, waiting for @cbrewer1689 response and fixed conditional button rendering logic

src/dataDisplay/personPanel/personPanelDetails.jsx
fantaJinMode: For avoiding regressions, I agree with @groberts314 of restoring the default value. Please provide your suggestion on this @cbrewer1689. According to that, I will make new changes

cbrewer1689: I think either way is good with me. I don't believe there would be any regressions with how this line current looks.

cbrewer1689 (approved):

src/dataDisplay/personPanel/personPanelDetails.jsx
groberts314: @fantaJinMode I've confirmed with @cbrewer1689 and _really do_ need to revert this change on the default value for `selectButtonProps`.  Because otherwise, callers that are not setting this prop (intending to accept the default value) will have their buttons mysteriously disappear! (because of the `!isEmpty(selectButtonProps)` check).

groberts314: Looking good as of 2ae8cb3.  I consider this resolved, but since we all participated, I'll keep the thread visible for now.

src/dataDisplay/personPanel/personPanelDetails.jsx
groberts314: This was resolved satisfactorily in c0008b4.

groberts314 (approved):

groberts314:
@cbrewer1689 Any objections to merging this and kicking out a 10.1.6 release?
I am happy to do that if you like.

PR #337, opened 2021-11-04
[dev] No PBI - Fixing Image type prop (it was always rendering 'user').

War777:
This will allow us to render `<Image type="person" />` (It the image was always rendering the `user` icon).

**Before:**

image

**After:**

image

### DIFF
--- a/src/dataDisplay/image/image.jsx
+++ b/src/dataDisplay/image/image.jsx
@@ -96,7 +96,7 @@ function Image(props) {
                     compact
                     size={avatarSize}
                     title="This person has no image"
-                    type="user"
+                    type={type}
                 />
             )}
         </ElementType>
--- a/src/dataDisplay/image/image.jsx
+++ b/src/dataDisplay/image/image.jsx
@@ -130,7 +130,7 @@ Image.defaultProps = {
     size: undefined,
     src: undefined,
     style: {},
-    type: undefined,
+    type: 'user',
 };
 
 export default Image;


groberts314 (approved):
LGTM
Curious ... why do we think this is `risk: high`?

src/dataDisplay/image/image.jsx
```diff
@@ -96,7 +96,7 @@ function Image(props) {
                     compact
                     size={avatarSize}
                     title="This person has no image"
-                    type="user"
+                    type={type}
```
cbrewer1689, line 99: I am thinking we should change the `defaultProp.type` to 'user'.

cbrewer1689 (changes requested):

src/dataDisplay/image/image.jsx
groberts314: Yeah, that might help minimize any possibility of regressions.  `undefined` as the default feels more correct to me in the general sense: i.e. not all Images are Avatar Images (in fact, most probably aren't), but, due to the preexisting bug where this was hard-coded to "user", making that default might be safer.

War777: No, it shouldn't have a default prop value because it'd override always the `src` prop in line 88.

```
<ElementType
        className={containerClasses}
        src={!type ? src : null}                        // <--------- here
        style={newStyle}
>
```

cbrewer1689: ... and I suppose the we're in a condition here where `type` is always going to be defined. But I think the original thinking here is that the icon should always be the user icon (image below). Is that a different case now?

image

War777: I checked the whole code and I think it works with a default prop so I defined it, thanks.

cbrewer1689: Okay, i am good with this then. Thank you for clarifying.

cbrewer1689 (approved):
PR looks good.

PR #340, opened 2021-11-10
[dev] PBI AB#58154: [Tech Debt] Add Box (MUI) To React-CM-UI

cbrewer1689:


docs/src/layout/box/box.jsx
```diff
@@ -0,0 +1,72 @@
...
+                    >
+                        {description}
+                    </Typography>
+
+                    <Heading
+                        anchorLink="address"
+                        variant="h2"
+                    >
+                        Box
+                    </Heading>
```
groberts314, line 53: Probably Line 49 should be `anchorLink="box"`

groberts314 (approved):
Looks good.  One small issue on docs

docs/src/layout/box/box.jsx
cbrewer1689: Fixed.

PR #343, opened 2021-12-02
[dev] PBI AB#59401: Upgrade React and React-DOM From v16 to v17

cbrewer1689:
The work in this PR helps HC to upgrade to React v17 in the future.


groberts314 (approved):
LGTM

Looks like you're actually updating us to 16.14, but added that peer dependency thing to allow other components referencing anywhere between 16.8.6 and 17 to work...?

Any gotchas we need to be worried?

cbrewer1689:
> Looks like you're actually updating us to 16.4, but added that peer dependency thing to allow other components referencing anywhere between 16.8.6 and 17 to work...?
>
> Any gotchas we need to be worried?

Correct, HC is on 16.14.x and really React-CM-UI should be on the same version. From my experience so far that upgrade doesn't regress anything.

I really don't believe there is any gotchas with this change, because really the only thing we're doing is syncing up our React version so we're dev'ing locally with the same version and then saying to the consumer install your own React version 16.8.6 or greater. These changes keep us safe from potential regressions in the future in my eyes.

groberts314:
Is anything like `componentWillUpdate()` and various other "unsafe" legacy stuff hard-deprecated?

cbrewer1689:
Not in 16 or v17. From the article [here](https://reactjs.org/blog/2020/10/20/react-v17.html), v17 is just a stepping stone to help with future upgrades to v18+.

fantaJinMode (approved):
Looks good to me.

PR #359, opened 2022-02-10
[dev] Bug AB#59185 [Mobile][My Ministries > Serving Opps] After adding a new Opportunity in My Ministry, the "Actions" pop out drawer does not close automatically | Extend Mobile Action Bar Actions Button and Actions Drawer to provide the option of toggling the Actions Drawer closed when an option is clicked

cletux87:
Execute onDrawerToggle instead of passing to parent component

### DIFF
--- a/src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
+++ b/src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
@@ -59,7 +59,8 @@ class ActionBarActionsButtonDrawerOption extends React.PureComponent {
                 }
             }
 
-            option.onClick(onDrawerToggle);
+            onDrawerToggle();
+            option.onClick();
             return false;
         }
 
--- a/src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
+++ b/src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
@@ -49,23 +49,22 @@ class ActionBarActionsButtonDrawerOption extends React.PureComponent {
 
         if (_.isFunction(option.onClick)) {
             if (option.disabled) {
-                return false;
+                return;
             }
 
             if (option.requiresPrompt) {
                 if (_.isFunction(onRequestPrompt)) {
                     onRequestPrompt(option);
-                    return false;
+                    return;
                 }
             }
 
             onDrawerToggle();
             option.onClick();
-            return false;
+            return;
         }
 
         onClick(isSelected ? {} : option);
-        return true;
     }
 
     render() {


groberts314 (approved):

groberts314:
This is needed for HC Admin 1.44.0 (per Proposed Release Version on [Bug AB#59185](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/59185).

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
```diff
@@ -59,7 +59,8 @@ class ActionBarActionsButtonDrawerOption extends React.PureComponent {
                 }
             }
 
-            option.onClick(onDrawerToggle);
+            onDrawerToggle();
+            option.onClick();
             return false;
```
groberts314, line 64: Are the Boolean return values actually meaningful?
If not, I would like to see them all changed to just `return;` where we need to bail out early, and then remove the final one on Line 68.

Thoughts?

_**Not a Blocker**_

groberts314 (approved):
Looks good to me.
One question/possible improvement with respect the `return`s in `onParentClick()` but I have designated it a NAB.

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
groberts314: Yeah, let's please do that, if you would be so kind.

cletux87: Sure

cbrewer1689: The change looks good, but are we waiting for return change before merging this?

cbrewer1689 (approved):

groberts314:
> are we waiting for return change before merging this?

If @cletux87 can bang that out pretty quickly, I think we should.

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
cletux87: Ready :)

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
```diff
@@ -49,22 +49,22 @@ class ActionBarActionsButtonDrawerOption extends React.PureComponent {
...
-                    return false;
+                    return;
                 }
             }
 
-            option.onClick(onDrawerToggle);
-            return false;
+            onDrawerToggle();
+            option.onClick();
+            return;
```
groberts314, line 64: We don't need this one.  Functions automagically return when we get to the end! =)

groberts314:
@cletux87 Looks good, although we technically we have one `return;` too many.  The last one can be elided.

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
groberts314: Actually nm.  This is not the end of the function.  Sorry; don't mind me.

groberts314:
@cbrewer1689 I say :shipit:

src/surfaces/actionBar/actionBarActionsButtonDrawerOption.jsx
cletux87: Not sure if I understand the return change @cbrewer1689  I can refactor the function to avoid using return. is that what you mean?

groberts314: @cletux87 I think @cbrewer1689 was referring to exactly the work you did in [e248d81](https://github.com/saddlebackdev/react-cm-ui/pull/359/commits/e248d81b3a627cf2528f88147ddfd9fb98c2a0e0).  So, it's done!  I think we can merge this if @cbrewer1689 agrees.

PR #365, opened 2022-03-01
[dev] PBI AB#59572: Removed Input's `showRequiredIndicator` component state

cbrewer1689:


src/inputs/input/input.jsx
```diff
@@ -226,10 +227,11 @@ const defaultProps = {
     style: null,
     tabIndex: null,
     type: null,
+    value: '',
```
groberts314, line 230: NICE!  This will get rid of the silly "switching back and forth between controlled and uncontrolled" issue, right?

groberts314 (approved):
LGTM

src/inputs/input/input.jsx
cbrewer1689: No, I don't believe this line disrupts that behavior.

groberts314: Oh, dang, I thought it might, since it was defaulting the value to empty string (which we usually have to do manually ourselves on the consumer side)

PR #367, opened 2022-03-14
[dev] Bug AB#68147 `<Checkbox>`es disabled with `disable` can be still be clicked

groberts314:
Defect with the `<Checkbox>` component noticed while troubleshooting the following:

**[Bug AB#68147](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/68147) | [Connection Cards > Follow Ups > Entries] - User is able to check and or uncheck disabled fields of questions in My follow Ups task's entry, in the Person Record drawer**

This fix is _NOT_ required the fix the HC Bug (which was addressed in https://github.com/saddlebackdev/church-management/pull/6085 by switching to use `disabled` instead of `disable`).  But it seems like something we _should_ fix.

### DIFF
--- a/src/inputs/checkbox/checkbox.jsx
+++ b/src/inputs/checkbox/checkbox.jsx
@@ -115,17 +115,24 @@ class Checkbox extends React.Component {
     }
 
     onClick(event) {
-        const { disabled, id, onChange } = this.props;
+        const {
+            disable,
+            disabled,
+            id,
+            onChange,
+        } = this.props;
+
         const { isChecked } = this.state;
 
-        if (!disabled) {
+        const isDisabled = disabled || disable;
+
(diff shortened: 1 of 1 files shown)


sircris (approved):
Looks good.

cbrewer1689 (approved):

cbrewer1689:
Question, @groberts314 are you okay with this going in 10.4.0? 10.4.0 would go out with HC 1.48.0.

groberts314:
> Question, @groberts314 are you okay with this going in 10.4.0? 10.4.0 would go out with HC 1.48.0.

Yep, that seems fine.  As indicated, this is not needed for a specific bug fix or anything, just a nice to have at some point.

PR #373, opened 2022-05-04
[dev] PBI 59401: tableRef Defensiveness 

cbrewer1689:


src/dataDisplay/table/table.jsx
```diff
@@ -401,13 +406,18 @@ class Table extends React.PureComponent {
             stickyColumnCount,
         } = this.props;
 
-        const stickyCells = this.tableRef.querySelectorAll(`.table-cell:nth-child(${stickyColumnCount})`);
+        console.log('this.tableRef', this.tableRef);
```
groberts314: We want to remove this before we merge, right?

groberts314 (approved):
Let's remove the `console.log()` and then it's ready to rosk 'n' roll!

src/dataDisplay/table/table.jsx
groberts314: Nice, gonezo!

PR #379, opened 2022-06-08
[dev] PBI AB#72763: Added The Ability Have A Custom Gradient On An Icon

cbrewer1689:
image


src/dataDisplay/icon/iconConstants.js
```diff
@@ -1,81 +1,6 @@
 import PropTypes from 'prop-types';
-import Utils from '../../utils/utils';
 
 export const TITLE_PROP_TYPE = PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.bool,
 ]);
```
groberts314, line 6: Is there still value in this?  Do we use it anywhere other than in `index.jsx`?  If not, we should consider removing it and just doing it inline with Icon's prop types declaration.
**_Not a Blocker_**

groberts314 (approved):
LGTM 👍 :shipit: 🚀

PR #380, opened 2022-06-08
[dev] No PBI - Added Button Types And Fixed Docs' Button API Page.

cbrewer1689:


cbrewer1689:
Depends on #379 so we can complete the Button's types to use TS enums.

groberts314 (approved):
LGTM 👍 :shipit: 🚀

src/inputs/button/button.jsx
```diff
@@ -2,25 +2,172 @@ import {
...
+     */
+    title: PropTypes.string,
+    /**
+     * Set transparent styles.
+     */
+    transparent: PropTypes.bool,
+    /**
+     * Set transparent styles.
+     */
+    type: PropTypes.oneOf(Object.values(Type)),
```
groberts314, line 132: Let's fix up this documentation comment (which appears to be copied and pasted) and maybe link to that MDN page about browser's button attribute.

src/inputs/button/models/type.model.ts
```diff
@@ -0,0 +1,7 @@
+enum Type {
```
groberts314, line 1: Let's rename to `ButtonType` oer our discussion

groberts314: Type renamed in 37f08fd.  Resolving,

src/inputs/button/button.jsx
groberts314: Resolved in a9ef7f5.

groberts314 (approved):
Approved latest changes including use of TS enums for Button props.  Love it!

PR #382, opened 2022-06-23
[dev] PBI AB#72773: Calmed Button Deprecated Warnings Down, FiltersDrawer Dropdown Value PropType Fix, TableHeaderCell Style Issue

cbrewer1689:
…rcell issue

### DIFF
--- a/src/inputs/dropdownButton/dropdownButton.jsx
+++ b/src/inputs/dropdownButton/dropdownButton.jsx
@@ -96,11 +96,14 @@ const propTypes = {
     /**
      * The type of the icon that sits to the left of the label.
      */
-    iconType: PropTypes.oneOf([
-        'caret-down',
-        'chevron-down',
-        'ellipsis-h',
-        'plus',
+    iconType: PropTypes.oneOfType([
+        PropTypes.bool,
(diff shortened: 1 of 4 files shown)


src/inputs/dropdownButton/dropdownButton.jsx
```diff
@@ -96,11 +96,14 @@ const propTypes = {
...
-        'ellipsis-h',
-        'plus',
+    iconType: PropTypes.oneOfType([
+        PropTypes.bool,
+        PropTypes.oneOf([
+            'caret-down',
+            'chevron-down',
+            'ellipsis-h',
+            'plus',
+        ]),
```
groberts314, line 106: NAB because it'd be a major breaking change.  But if we're also allowing a Boolean, then IMO `icon` would be a better name than `iconType`.  i.e. `icon` (`icon={true}`) or `icon={false}` reads better than `iconType={false}`.

cbrewer1689: I totally agree and the reason for not making the change is because as you said would be breaking. Kinda of a bummer.

PR #407, opened 2022-11-22
[dev] No PBI - Docs updates, Import examples.

War777:
This PR will:

- Update the react-cm-ui package import in examples to:
```

import {
    ...
} from '@saddlebackchurch/react-cm-ui';
```

- Add the `inputMode` prop in the `<Input />` component so mobile browsers can have a better interaction when handling only numbers for example (introduced because is needed for the Giving MVP custom amount input field).


src/surfaces/card/cardBase.tsx
```diff
@@ -9,7 +9,7 @@ type PropTypes = {
      * or the new Card component.
      */
     designVersion?: number;
-};
+};}
```
groberts314: This doesn't look right

groberts314 (approved):

src/surfaces/card/cardBase.tsx
War777: Fixed, thanks for the catch.

groberts314:
@War777 @raoufr Assuming that we want a NPM package release right way (once this is approved and merged), I think 10.6.1 (just bumping patch segment of the semver) is appropriate, as the actual change to the functionality of the main library is pretty modest.

raoufr:
> @War777 @raoufr Assuming that we want a NPM package release right way (once this is approved and merged), I think 10.6.1 (just bumping patch segment of the semver) is appropriate, as the actual change to the functionality of the main library is pretty modest.

Agreed, thank very much @groberts314

raoufr (approved):
Looking good to me

PR #438, opened 2023-05-09
[dev] AB#92442 [React Library] Refactor UI Library for Dropdown Menus and Dropdown Buttons

mkemper86:
This fixes a bug in allowing the `DropdownMenu` inside the `DropdownButton` to automatically close when the page is clicked outside of the button. It also adds a couple of props to allow overriding the horizontal positioning of the `DropdownMenu` to accommodate some updated designs that have right-aligned menus.

### DIFF
--- a/src/inputs/dropdownMenu/dropdownMenu.jsx
+++ b/src/inputs/dropdownMenu/dropdownMenu.jsx
@@ -2,6 +2,7 @@ import {
     debounce,
     get,
     isFunction,
+    isNil,
 } from 'lodash';
 import ClassNames from 'classnames';
 import PropTypes from 'prop-types';
@@ -15,9 +16,10 @@ import {
 } from './dropdownMenuConstants';
 import makeStyles from '../../styles/makeStyles';
 
-const propTypes = {
+export const propTypes = {
     children: PropTypes.node.isRequired,
     className: PropTypes.string,
+    horizontalAlign: PropTypes.string,
     id: PropTypes.string,
     isOpen: PropTypes.bool.isRequired,
     getParentContainer: PropTypes.func,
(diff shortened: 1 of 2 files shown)


src/inputs/dropdownMenu/dropdownMenu.jsx
```diff
@@ -230,7 +245,7 @@ function DropdownMenu(props) {
         onDropdownMenuReposition();
 
         return () => {
-            if (isOpen) {
+            if (!isOpen) {
```
groberts314, line 248: Is this change intentional?  Can I understand why?

groberts314 (commented):
Mostly looks okay.  Just one question.

src/inputs/dropdownMenu/dropdownMenu.jsx
mkemper86: Yes it's intentional. We want the listener events to be active when the menu is _open_ because we want to be able to close the menu by clicking on the document outside the dropdown. If the events are removed when the menu is open, then the `onClickOutside` will never get run, right?

groberts314: Yeah, you're making sense.  Like we want to remove the events when it is closed.  I wonder was this just a long-running bug?

groberts314 (approved):

src/inputs/dropdownMenu/dropdownMenu.jsx
mkemper86: I think so...the deprecated `DropdownMenu` works correctly with similar code, so I think it was just an oversight.

War777 (approved):

PR #440, opened 2023-06-22
[dev] [No Bug\PBI] Prevent crash in Dropdown Menu component

groberts314:
Add defensive check on `parentContainer` before de-referencing it to call `contains()` method in `onClickOutside()` function, which can cause a runtime error if it's nullish.

image

image

### DIFF
--- a/src/inputs/dropdownMenu/dropdownMenu.jsx
+++ b/src/inputs/dropdownMenu/dropdownMenu.jsx
@@ -224,7 +224,7 @@ function DropdownMenu(props) {
     useEffect(() => {
         function onClickOutside(event) {
             const parentContainer = dropdownMenuRef.current;
-            const containsTarget = parentContainer.contains(event.target);
+            const containsTarget = parentContainer?.contains(event.target) ?? false;
 
             if (
                 !containsTarget ||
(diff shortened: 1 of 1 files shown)


KacosPro (approved):

raoufr (approved):

src/inputs/dropdownMenu/dropdownMenu.jsx
```diff
@@ -224,7 +224,7 @@ function DropdownMenu(props) {
     useEffect(() => {
         function onClickOutside(event) {
             const parentContainer = dropdownMenuRef.current;
-            const containsTarget = parentContainer.contains(event.target);
+            const containsTarget = !isNil(parentContainer) && parentContainer.contains(event.target);
```
ajaybushel, line 227: Just an idea:

We can minimise the dependency on Lodash, we can do:

```parentContainer?.contains?.(event.target)```

In any case the returned value would be `undefined` or `null` if `parentContainer` is empty.

groberts314: @ajaybushel Yes we can, but it required installing additional Babel plugins not currently installed here, and we were in a little bit of a hurry with this release.  But just so you know I had the same thought, see 37bac18 and then 84d72d7.  GMTA! =)

PR #461, opened 2023-08-31
[dev] React Select V5 Based - First Cut.

War777:
This PR introduces the first iteration and basic implementaion for the React Select component using V5 (previous one is still V1).

- This will allow us to work with Modals without list problems.
- It should be imported as `SelectNext` from `@saddlebackchurch/react-cm-ui`
- There shouldn't be regressions issues with the previous `Select` as we're not removing it (we're still maintaining the V1 dependency in `react-cm-ui`)

We should gradually add the rest of the `Select` use cases, this cut purpose is to fix a couple of Finance blocking issues.

- [BUG AB#96325](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/96325) - [Finance > Add Sub Account] - The "Names" of the codes with longer characters are cut off and not fully displayed in the drop-down menus, of the "Add Sub Account" modal
- [BUG AB#95178](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/95178) - [Finance > Add Account] - User should not be able to scroll up and down, in the "Add Account" modal (dialog box) 

image


groberts314:
This seems a big enough change that it warrents bumping the minor IMO.  So ... React CM UI 10.7.0 baby! 🚀

src/inputs/selectNext/selectNext.tsx
```diff
@@ -0,0 +1,439 @@
...
+    label: string;
+    /**
+     * Whether the menu should use a portal, and where it should attach
+     * An example can be found in the Portaling documentation https://react-select.com/advanced#portaling
+     */
+    menuPortalTarget?: HTMLElement | null;
+    /**
+     * Text to display when there are no options
+     */
+    noOptionsMessage?: () => string | null;
```
groberts314, line 54: Should also allow just straight up strings, as well as functions that produce strings?
NAB

groberts314: Don't know if we need `| null`.  Like it should either be a thing we expect or `undefined` (prop omitted), right?
Also NAB.

src/inputs/selectNext/selectNext.tsx
```diff
@@ -0,0 +1,439 @@
...
+    /**
+     * Handle change events on the select
+     */
+    onChange: (nextValue: { value: string; label: string}) => void;
```
groberts314, line 58: We can use `Option` instead of `{ value: string; label: string}`, right?

src/inputs/selectNext/selectNext.tsx
```diff
@@ -0,0 +1,439 @@
...
+    value: string;
+    label: string;
+};
+
+/**
+ * The Select component represents a control that provides a menu of options.
+ */
+// eslint-disable-next-line prefer-arrow-callback
+const SelectNext = React.forwardRef(function SelectNext(
+    props: PropTypes, ref: LegacyRef<HTMLDivElement>,
```
groberts314, line 355: Shouldn't these be on separate lines? =)

src/inputs/selectNext/selectNext.tsx
```diff
@@ -0,0 +1,439 @@
...
+                    DropdownIndicator: CustomArrow,
+                    MenuList: CustomMenuList,
+                    Option: CustomOption,
+                }}
+                // @ts-ignore
+                dropdownMenuMaxHeight={dropdownMenuMaxHeight}
+                dropdownMenuMinHeight={dropdownMenuMinHeight}
+                isSearchable={isSearchable}
+                menuPortalTarget={menuPortalTarget}
+                name="cm_ui_select_next"
```
groberts314, line 420: Do we want to hard-code a name?  That needs to be settable via prop, no?
Like if consumer has a form with multiple selects, don't they need distinct names?

groberts314 (commented):
A few mostly minor concerns

src/inputs/selectNext/selectNext.tsx
War777: That's how it's declared in the Select propTypes 

image

src/inputs/selectNext/selectNext.tsx
War777: You're right!

src/inputs/selectNext/selectNext.tsx
War777: Done!

src/inputs/selectNext/selectNext.tsx
groberts314: Can we just do `noOptionsMessage?: () => string;`?  Leave off `| null`?

src/inputs/selectNext/selectNext.tsx
War777: Defined via props!

src/inputs/selectNext/selectNext.tsx
groberts314: Looking better as of e03236f: https://github.com/saddlebackdev/react-cm-ui/pull/461/commits/e03236f6c3d711887b1ac186fb03b4b86c1fc02f#diff-a86b373e2035b10f8f8a9550a5f1f02cc22e5be10743d74e35732aaaa88295b7L54-R58

Resolving.

src/inputs/selectNext/selectNext.tsx
groberts314: Looks good.  Resolving.

src/inputs/selectNext/selectNext.tsx
groberts314: Gracias!  Resolving.

groberts314 (approved):

groberts314:
@War777 does this need any additional testing or anything?  Or should I go ahead and roll it?

War777:
> @War777 does this need any additional testing or anything? Or should I go ahead and roll it?

Select tests are part of the original project and as we're extending only styling I guess we can merge it as it is, we'll add tests as we go implementing custom stuff from our side, thanks.

PR #466, opened 2023-10-12
[dev] BUG AB#102385 - [Finance > Segments & Codes > Edit Code] - User should be able to nullify the "Associated HC Campus" (if they want to) of a Campus code, in the drop down menu, as it is not a required field

War777:
This PR introduces the `clearable` option for the `<SelectNext />` component so it can fix the following requirement:
- [BUG AB#102385](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/102385) - [Finance > Segments & Codes > Edit Code] - User should be able to nullify the "Associated HC Campus" (if they want to) of a Campus code, in the drop down menu, as it is not a required field.

image

### DIFF
--- a/src/inputs/selectNext/selectNext.tsx
+++ b/src/inputs/selectNext/selectNext.tsx
@@ -1,7 +1,7 @@
 import ClassNames from 'classnames';
 import {
-    isFunction,
     isEmpty,
+    isFunction,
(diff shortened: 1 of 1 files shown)


raoufr (approved):
Looking good to me

PR #476, opened 2023-11-06
[hotfix/10.7.3-hotfix/1] AB#102944 -- add new  icons for sections in the Journey Builder

IamNotHuman:
added Christmas  icons


groberts314 (changes requested):
Let's please get rid of all encoded whitespace characters and string concatenations, and just put the entire string on one line (like the other existing icons in this file)

groberts314 (approved):

PR #478, opened 2023-11-16
[dev] PBI AB#103112 [Data Capture - Connection Form - Entries] Ability to Read a Sensitive Connection Question Answer

IlyaRadinsky:
- Fixed margin bugs for `<Checkbox/>`, `<Radio />`, `<Input />`
- Fixed the case when `<Input />` has an icon and `isRedacted` flag at the same time
- Added supporting `isRedacted` flag for the `<DatePickerInput />`

### DIFF
--- a/src/inputs/datePickerInput/datePickerInput.jsx
+++ b/src/inputs/datePickerInput/datePickerInput.jsx
@@ -129,6 +129,12 @@ const propTypes = {
     includeDates: PropTypes.arrayOf(MomentPropTypes.momentObj),
 
     /**
+     * To prevent sensitive data from being read, we need to be able to block the contents of the
+     * control with a gray placeholder. This flag triggers this kind of display instead of the usual one.
+     */
+    isRedacted: PropTypes.bool,
+
+    /**
      * Specifies a label for the DatePickerInput control.
      */
     label: PropTypes.string,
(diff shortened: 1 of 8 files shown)


src/inputs/datePickerInput/datePickerInput.jsx
```diff
@@ -617,10 +625,11 @@ class DatePickerInput extends React.PureComponent {
                                         color={iconColor}
                                         compact
                                         disable={isDisabled}
-                                        onClick={this.onIconClick}
+                                        onClick={isRedacted ? null : this.onIconClick}
```
groberts314, line 628: NAB: Is `null` a good value here?  Might `undefined` or `noop` be better?
NAB and discretionary if this is working code.
JS is _so confusing_ with both `undefined` and `null` ... sigh.

groberts314 (approved):

src/inputs/datePickerInput/datePickerInput.jsx
IlyaRadinsky: there should be some kind of nil value, not `noop`

groberts314: This is probably fine as-is.  Resolving.

PR #479, opened 2023-11-16
[dev] Bug AB#103415: [Event Central][Paid Events 2.0] - An additional "0" (zero) will be added automatically (at the beginning of the Price of the Ticket amount), when user tries to Edit the Ticket's Regular price field

romantech48:
[Bug 103415](https://dev.azure.com/saddlebackchurch/Church%20Management/_workitems/edit/103415): [Event Central][Paid Events 2.0] - An additional "0" (zero) will be added automatically (at the beginning of the Price of the Ticket amount), when user tries to Edit the Ticket's Regular price field

- Added option to ignore value enforcement by `required` for `number` input
- Fixed issue with `max` not being respected

### DIFF
--- a/src/inputs/input/input.jsx
+++ b/src/inputs/input/input.jsx
@@ -25,6 +25,24 @@ const propTypes = {
      */
     allowDecimals: PropTypes.bool,
     /**
+     * Determines whether or not the field allows a completely empty value
+     * for Inputs of type "number", when the Input is also marked `required`.
+     *
+     * This is purely a UI/UX/behavioral option that does not really change the
+     * 'required field' semantics.  By default, an Input of type number that is
+     * required will default to either the configured minimum value (specified via the `min` prop)
+     * or else 0.  By setting this prop to `true`, you can suppress that behavior and allow
+     * the input to be completely empty.  The use case for this is for cases where the user wishes
+     * to completely clear out the existing value and type a new one
+     * (versus having some default automatically entered) for slightly smoother UX.
+     */
+    allowEmptyValueWithRequired: PropTypes.bool,
+    /**
+     * Indicates whether or not the field allows negative numbers (or just positive numbers)
+     * for Input of type "number".
+     */
+    allowNegativeNumbers: PropTypes.bool,
+    /**
      * Forces the Input component to always show the required indicator
      * next to the label. The default behavior (if this prop is omitted or false) is for
      * the required field indicator to disappear once a value has been entered.
@@ -34,16 +52,6 @@ const propTypes = {
      * Gives Input immediate focus.
      */
     autoFocus: PropTypes.bool,
-    /**
-     * Indicates whether or not the field allows negative numbers (or just positive numbers)
-     * for Input of type "number".
-     */
-    allowNegativeNumbers: PropTypes.bool,
-    /**
-     * Indicates whether or not the field allows empty value
-     * for Input of type "number".
-     */
-    allowEmptyValueWithRequired: PropTypes.bool,
     /**
      * Additional classes.
      */
@@ -227,11 +235,11 @@ const propTypes = {
 };
 
 const defaultProps = {
-    autoComplete: null,
     allowDecimals: true,
-    allowNegativeNumbers: true,
     allowEmptyValueWithRequired: false,
+    allowNegativeNumbers: true,
     alwaysShowRequiredIndicator: false,
+    autoComplete: null,
     autoFocus: null,
     className: null,
     dataTestId: undefined,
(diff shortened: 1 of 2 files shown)


src/inputs/input/input.jsx
```diff
@@ -39,6 +39,11 @@ const propTypes = {
      * for Input of type "number".
      */
     allowNegativeNumbers: PropTypes.bool,
+    /**
+     * Indicates whether or not the field allows empty value
+     * for Input of type "number".
+     */
+    allowEmptyValueWithRequired: PropTypes.bool,
```
groberts314, line 46: I think I understand what we're wanting to do here, but the name of this prop is maybe a bit confusing.  This name makes it sound like you can opt out of something that's required.  Maybe the name is okay (or, at least, I am having trouble thinking of a better one off the top of my head) but we just need to add some more explanation to the comment.  I can maybe take a crack at that.

src/inputs/input/input.jsx
```diff
@@ -368,10 +375,10 @@ class Input extends React.PureComponent {
                             }
                         }
                     } else {
-                        const newValueAsNumber = +newValue;
+                        let newValueAsNumber = +newValue;
 
                         if (isNumber(max)) {
-                            newValue = Math.min(max, newValueAsNumber);
+                            newValueAsNumber = Math.min(max, newValueAsNumber);
```
groberts314, line 381: Help me understand this change please.  Is it not necessary to update `newValue` as well (as the original code had it), since that is the value that will actually be used to update state?  Like maybe we should update both `newValueAsNumber` and `newValue`?

groberts314 (changes requested):
I have a few questions/concerns on this one.

src/inputs/input/input.jsx
romantech48: The original implementation ignored the max ceiling. It's went like this (isNumber check omitted):
```
1. newValuesAsNumber = +newValue; // 999
2. newValue = Math.min(max, newValueAsNumber); 
// ...min(100, 999) > newValue = 100
2. newValue = Math.max(min, newValueAsNumber); 
// ...max(10, 999) > newValue = 999 (overwritten)
```
So we need to update newValueAsNumber. We could use newValue instead of newValuesAsNumber in max func, if needed.

src/inputs/input/input.jsx
romantech48: Maybe something like this: 

> Overrides the default behavior of enforcing a value for Input of type "number" when "required" is specified.

Or instead of introducing a new prop. We can adjust the way how `alwaysShowRequiredIndicator` works (to show it unconditionally). As currently:

> Forces the Input component to always show the required indicator next to the label. The default behavior (if this prop is omitted or false) is for the required field indicator to disappear once a value has been entered.

groberts314: Hmm.  What we're really doing though is allowing it to be completely empty, as opposed to defaulting to the Minimum value, right?  I guess that's how I understood the behavior.

Like the bug you're trying to fix says "I want to be able to backspace and delete the entire value in a number input (for the ticket price), but it doesn't let me do that because it always insists on putting a 0 there.", right?

And you're fix is "hey, here's this new prop that will override that behavior and not force the minimum value".

Is that a correct understanding?

src/inputs/input/input.jsx
groberts314: What if we did this?
```JavaScript
newValue = newValueAsNumber = Math.min(max, newValueAsNumber);
```

src/inputs/input/input.jsx
romantech48: @groberts314 Yes, the new prop is optional and should not break existing code. Is the description I proposed is suitable or do we need a different approach?

groberts314: @romantech48 Sorry for the delay here.  I will try and propose something and/or change it myself later today.  I do want to get this enhancement merged in and released (along with #481) as version `10.9.2`.

My main concern is that we thoroughly explain that using this new prop makes a small change to UI/UX (suppressing the default to the minimum value and instead allowing the input to be empty), without altering the 'required' semantics (i.e. it is still a required field and the user is obligated to fill a new value after completely removing the old one to satisfy the requirement).

groberts314: Updated the comment to my satisfaction: https://github.com/saddlebackdev/react-cm-ui/pull/479/commits/c871e0933ad89faafb702642f651acd191d23eb7#diff-6091a1e95f11eb778b4baba032cd2b4ae577f1410727154a749e55c473f3a3d8R27-R39

Set it as follows:
> Determines whether or not the field allows a completely empty value for Inputs of type "number", when the Input is also marked `required`.
>
> This is purely a UI/UX/behavioral option that does not really change the 'required field' semantics.  By default, an Input of type number that is required will default to either the configured minimum value (specified via the `min` prop) or else 0.  By setting this prop to `true`, you can suppress that behavior and allow the input to be completely empty.  The use case for this is for cases where the user wishes to completely clear out the existing value and type a new one (versus having some default automatically entered) for slightly smoother UX.

groberts314 (approved):

PR #481, opened 2023-11-22
[dev] AB#101313 - Added isFiltering in filterOption for filter rails

fantaJinMode:
PR contains:
- Introducing `isFiltering` in `filterOption` for filter rails to make better UX experience for `Clear All` button

### DIFF
--- a/src/layout/page/pageFiltersRail.jsx
+++ b/src/layout/page/pageFiltersRail.jsx
@@ -14,6 +14,10 @@ const propTypes = {
          */
         isDirty: PropTypes.bool.isRequired,
         /**
+         * If `true`, filters are applied
+         */
+        isFiltering: PropTypes.bool.isRequired,
+        /**
          * Apply filters Button onClick event handler
          */
         onApply: PropTypes.func.isRequired,
--- a/src/surfaces/filtersRail/filtersRail.jsx
+++ b/src/surfaces/filtersRail/filtersRail.jsx
@@ -40,9 +40,13 @@ const propTypes = {
      */
     filterOptions: PropTypes.shape({
         /**
-         * If `true`, filters are applied
+         * If `true`, filters are dirty
          */
         isDirty: PropTypes.bool.isRequired,
+        /**
+         * If `true`, filters are applied
+         */
+        isFiltering: PropTypes.bool.isRequired,
         /**
          * Apply filters Button onClick event handler
          */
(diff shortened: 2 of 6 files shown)


src/layout/page/pageFiltersRail.jsx
```diff
@@ -13,6 +13,10 @@ const propTypes = {
          * If `true`, filters are applied
          */
         isDirty: PropTypes.bool.isRequired,
```
groberts314, line 15: Should the comment for `isDirty` be changed?
Maybe something like:
> If `true`, then filter options have been modified but not yet applied to the list results.

src/surfaces/filtersRail/filtersRail.jsx
```diff
@@ -285,7 +290,7 @@ function FiltersRail(props) {
                             <div className={actionButtonsClasses}>
                                 <A
                                     className={clearFiltersClasses}
-                                    disable={!filterOptions.isDirty}
+                                    disable={!filterOptions.isDirty || !filterOptions.isFiltering}
```
groberts314, line 293: Is OR correct here?  Or should it be AND?
Also, can we use `disabled` rather than `disable`, or does `<A>` component only use `disable`?

groberts314 (changes requested):
A couple of small concerns

src/surfaces/filtersRail/filtersRail.jsx
fantaJinMode: `OR` is the correct condition applied. `<A>` component only use `disable` 😞

src/layout/page/pageFiltersRail.jsx
fantaJinMode: Fixed. [Commit Link](https://github.com/saddlebackdev/react-cm-ui/pull/481/commits/cad8178f27e4cf0a81c6d50d27fdc4f7365e3199)

src/surfaces/filtersRail/filtersRail.jsx
groberts314: Got it.  Thanks for confirming.  I think it's so many NOTs, coupled with the fact that it's the condition for with the Link Button should NOT be active (i.e. should be disabled) -- that what made it slightly tricky for me. =)  But I think it makes sense.

src/layout/page/pageFiltersRail.jsx
groberts314: Thanks.  Resolving.

groberts314 (approved):

PR #490, opened 2024-03-29
[dev] [No PBI] Select onInputChange Prop

razmik-medoxi:


groberts314:
@razmik-medoxi What functionality does this enable?

src/inputs/select/select.jsx
```diff
@@ -192,6 +196,7 @@ const defaultProps = {
     noResultsText: 'No results found',
     onChange: null,
     onClose: undefined,
+    onInputChange: null,
```
groberts314, line 199: NAB: `undefined` is probably better here than `null`, not that it matters much.
I guess I like `null` as a default for _data_ (like this prop can be some specific object/type, or `null`), but for things like optional CSS classes or styles or event handler functions, I think `undefined` is appropriate.

razmik-medoxi:
> @razmik-medoxi What functionality does this enable?

@groberts314 I'm exposing react-select's `onInputChange` prop, I need to get select's input value to trigger an API request and get the events list

groberts314:
> I'm exposing react-select's onInputChange prop, I need to get select's input value to trigger an API request and get the events list

How is that different than the ordinary `onChange` event?

razmik-medoxi:
> > I'm exposing react-select's onInputChange prop, I need to get select's input value to trigger an API request and get the events list
> 
> How is that different than the ordinary `onChange` event?

`onChange` is for selectedOption changes, this one is for the input changes

groberts314:
> `onChange` is for selectedOption changes, this one is for the input changes

So, this event fires if someone types in the input versus picking something from the dropdown menu?

razmik-medoxi:
> > `onChange` is for selectedOption changes, this one is for the input changes
> 
> So, this event fires if someone types in the input versus picking something from the dropdown menu?

exactly

groberts314:
@razmik-medoxi These changes are available in [`@saddlebackchurch/react-cm-ui`](https://www.npmjs.com/package/@saddlebackchurch/react-cm-ui) version **`10.9.7`**.

PR #502, opened 2024-08-01
[dev] PBI AB#109601 [Back Office][Digital Program > Builder] - Missing Symbol Icons need to be added and Duplicates to be removed

IlyaRadinsky:


groberts314 (changes requested):
Docs need to be updated as well.
See: https://github.com/saddlebackdev/react-cm-ui/blob/features/fix-cm-ui-docs/docs/src/dataDisplay/icon/icon.jsx#L39
New icons need to be added there somewhere.

groberts314 (approved):

